PR #19, opened 2020-09-04
SAAS-283 basic platform account

atymchuk:
This PR features:
* private and public routes
* two layouts, both responsive
* a redux store
* login async actions
* logout async actions
* a test container

TODO: add signup actions, make gRPC calls to BE

NOTE: the password MUST contain at least 10 characters (if not more, sync with BE)


nicolalamacchia (approved):

packages/platform-ui/src/pages/Login/Login.tsx
```diff
@@ -1,58 +1,70 @@
...
-  const styles = getLoginStyles(theme);
+  const styles = useStyles(getStyles);
+  const history = useHistory();
+
+  const handleLoginSubmit = useCallback(
+    async (credentials: Credentials) => {
+      try {
+        store.dispatch(authLoginAction.request(credentials));
+        await sleep();
+        store.dispatch(authLoginAction.success());
```
tiagomotasantos, line 29: I think success and failure actions should be dispatched at the end of the request action and not here. I know there's no BE integration yet but wanted to leave this here anyway.

tiagomotasantos (changes requested):

packages/platform-ui/src/core/messages.ts
```diff
@@ -0,0 +1,6 @@
+export const Messages = {
+  companyName: 'Percona',
+  productDescription:
+    'Single pane of glass for managing and monitoring the performance of your MySQL, MariaDB, PostgreSQLand MongoDB databases.',
```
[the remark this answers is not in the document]
nicolalamacchia: Done

packages/platform-ui/src/pages/Login/Login.tsx
atymchuk: That's quick and dirty. Of course, we will not leave it forever like this )

tiagomotasantos (dismissed):

packages/platform-ui/src/core/constants.ts
```diff
@@ -4,4 +4,4 @@ export const PRIVACY_POLICY_URL = 'https://per.co.na/pmm/platform-privacy';
 
 export const PASSWORD_MIN_LENGTH = 10;
 
-export const PLATFORM_AUTH_API_BASE_URL = process.env.NODE_ENV === 'development' ? 'https://platform-dev.percona.com' : 'https://platform.percona.com';
+export const PLATFORM_AUTH_API_BASE_URL = process.env.REACT_APP_PLATFORM_AUTH_API_BASE_URL ?? '';
```
atymchuk, line 7: Actually, we don't need coalescing here.
```suggestion
export const PLATFORM_AUTH_API_BASE_URL = process.env.REACT_APP_PLATFORM_AUTH_API_BASE_URL;
```

packages/platform-ui/src/components/Root/Main.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <LoginPage />
+        </PublicRoute>
+        <PublicRoute path={Routes.signup}>
+          <SignupPage />
+        </PublicRoute>
+        <Route path="*">
+          <NotFound />
+        </Route>
+      </Switch>
+      <ToastContainer
```
atymchuk, line 50: Probably not the best place for the ToastContainer, I'd move it to index.tsx

nicolalamacchia: Done

packages/platform-ui/src/core/constants.ts
nicolalamacchia: Why? I think it's needed, since the value can be either string or undefined, so it's either that or the exclamation mark? Otherwise TS will complain about that (and I would prefer solve this on one line, instead of on each line referring to this variable) :)

atymchuk: Definitely exclamation mark. What is the reason we want to pass an empty string to the API?

packages/platform-ui/src/components/Authenticated/Authenticated.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <section className={styles.container}>
+        <p>{Messages.authenticatedFirstLine}</p>
+        <p>{Messages.authenticatedSecondLine}</p>
+        <p>
+          <em>
+            <b>{email}</b>
+          </em>
+        </p>
+          <LinkButton className={styles.downloadPMMButton} href={DOWNLOAD_PMM_LINK}>
+            Download PMM
```
tiagomotasantos, line 47: Move to messages file

tiagomotasantos (changes requested):

packages/platform-ui/src/components/Authenticated/Authenticated.tsx
nicolalamacchia: Good catch

tiagomotasantos (approved):

tiagomotasantos (dismissed):

nicolalamacchia (approved):

AlekSi (approved):

PR #24, opened 2020-09-14
SAAS-340 cypress integration

yurkovychv:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,69 @@
+
```
atymchuk, line 1: Consider renaming this file to `e2e.yml` - much more explicit.

packages/platform-ui/cypress.json
```diff
@@ -0,0 +1,24 @@
...
+  "defaultCommandTimeout": 8000,
+  "execTimeout": 60000,
+  "pageLoadTimeout": 60000,
+  "requestTimeout": 15000,
+  "responseTimeout": 30000,
+  "video": true,
+  "videoUploadOnPasses": false,
+  "failOnStatusCode": false,
+  "ignoreTestFiles": ["*utils.js", "locators.js"],
+  "baseUrl": "http://localhost:3000/",
```
atymchuk, line 17: That trailing slash may not be needed. Pls refer to https://docs.cypress.io/guides/references/best-practices.html#Configuration-file-cypress-json-by-default

packages/platform-ui/cypress/fixtures/login.json
```diff
@@ -0,0 +1,4 @@
+{
```
atymchuk, line 1: Does this have to be committed?

yurkovychv: not for now actually, I removed it

packages/platform-ui/cypress.json
yurkovychv: done, thanks

.github/workflows/tests.yml
yurkovychv: done, thanks

packages/platform-ui/cypress/support/stub.services.js
```diff
@@ -1,17 +1,28 @@
...
     headers: {
-      'set-cookie': '__Host-PP-Session=102ZX6eUv6mQ4qgEQSaIGIupQ; Path=/; Expires=Tue, 22 Sep 2021 08:08:05 GMT; Max-Age=604799; HttpOnly; Secure; SameSite=Strict',
-      "content-type": "application/grpc-web-text",
-      'status': 200,
-      'vary': origin
+      'set-cookie':
+        '__Host-PP-Session=102ZX6eUv6mQ4qgEQSaIGIupQ; Path=/; Expires=Tue, 22 Sep 2021 08:08:05 GMT; Max-Age=604799; HttpOnly; Secure; SameSite=Strict',
+      'content-type': 'application/grpc-web-text',
+      status: 200,
+      vary: origin,
```
atymchuk, line 14: Here `origin` seems to be a variable. Where should its value come from? If it is `window.origin`, then we should be explicit to avoid guessing.

packages/platform-ui/cypress/integration/auth/constants.js
```diff
@@ -0,0 +1,5 @@
+export const TERMS_OF_SERVICE_URL = 'https://per.co.na/pmm/platform-terms';
```
atymchuk, line 1: There is no need to copy those values, we could as well import them from where they are initially defined.

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,65 @@
...
+  if (test.state === 'failed') {
+    let item = runnable;
+    const nameParts = [runnable.title];
+
+    while (item.parent) {
+      nameParts.unshift(item.parent.title);
+      item = item.parent
+    }
+    if (runnable.hookName) {
+      nameParts.push(`${runnable.hookName} hook`)
```
atymchuk, line 46:
```suggestion
      nameParts.push(`${runnable.hookName} hook`);
```

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+Cypress.on('test:after:run', (test, runnable) => {
+
+  if (test.state === 'failed') {
+    let item = runnable;
+    const nameParts = [runnable.title];
+
+    while (item.parent) {
+      nameParts.unshift(item.parent.title);
+      item = item.parent
```
atymchuk, line 43: I see lots of code style violations here :|

atymchuk, line 43:
```suggestion
      item = item.parent;
```

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,65 @@
...
+      nameParts.push(`${runnable.hookName} hook`)
+    }
+
+    const fullTestName = nameParts
+        .filter(Boolean)
+        .join(' -- ');
+
+    const imageUrl = `screenshots/${
+        Cypress.spec.name
+    }/${fullTestName} (failed).png`;
```
atymchuk, line 55: A space in a file name? Consider an underscore

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,65 @@
...
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
```
atymchuk, line 17:
```suggestion
import './commands';
```

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,65 @@
...
+import addContext from 'mochawesome/addContext'
```
atymchuk, line 18:
```suggestion
import addContext from 'mochawesome/addContext';
```

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,65 @@
...
+import setup from 'cypress-cy-select';
+import {configure} from '@testing-library/cypress'
```
atymchuk, line 20:
```suggestion
import {configure} from '@testing-library/cypress';
```

packages/platform-ui/cypress/support/index.js
yurkovychv: actually the final screenshot name will be something like: Login -- should be able to see the login form (failed).png

packages/platform-ui/cypress/integration/auth/constants.js
yurkovychv: The intention was to make tests constants separated from the FE constants. In order to track changes and catch bugs

packages/platform-ui/cypress/support/stub.services.js
yurkovychv: looks like we do not need it here, removed

packages/platform-ui/cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
...
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
+ */
+module.exports = (on, config) => {
+  // `on` is used to hook into various events Cypress emits
+  // `config` is the resolved Cypress config
+}
+
```
nicolalamacchia, line 22: This empty file is just an example, we shouldn't commit it now

packages/platform-ui/cypress/support/commands.js
```diff
@@ -0,0 +1,32 @@
+// ***********************************************
+// This example commands.js shows you how to
+// create various custom commands and overwrite
+// existing commands.
```
nicolalamacchia, line 4: I would remove example lines and lines stating that a file is an example from these files

packages/platform-ui/cypress/support/index.js
nicolalamacchia: This is marked as resolved, but still missing the semicolon. Same for the following lines.

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,65 @@
...
+      nameParts.unshift(item.parent.title);
+      item = item.parent
+    }
+    if (runnable.hookName) {
+      nameParts.push(`${runnable.hookName} hook`)
+    }
+
+    const fullTestName = nameParts
+        .filter(Boolean)
+        .join(' -- ');
```
nicolalamacchia: Please, add a comment explaining what this code does

nicolalamacchia: This whole block of code is pretty obscure

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,74 @@
...
+        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+      - name: Start project and execute tests
+        uses: percona-platform/github-action@v2
+        with:
+          working-directory: packages/platform-ui
+          browser: chrome
+          start: yarn start
+          wait-on: 'http://localhost:3000'
+          wait-on-timeout: 120
```
nicolalamacchia, line 62: Note that this is not necessarily the full project. We could end up having more than one package in the future. How would this scale?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Build Core
+        run: lerna run build --scope='@percona/platform-core'
+
+      - name: Link
+        run: lerna link
+
+      # The solution was taken from https://github.com/expo/expo-github-action/issues/20#issuecomment-541771861
+      # in order to avoid file watchers limit errors
+      - name: Increase watchers
+        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```
nicolalamacchia, line 53: Please, give more details about the problem this solution is for

nicolalamacchia (changes requested):

.github/workflows/e2e.yml
yurkovychv: `chokidar` was throwing `Error: ENOSPC: System limit for number of file watchers reached` during project startup in the workflow. For example: [in this workflow execution](https://github.com/percona-platform/saas-ui/runs/1116292311?check_suite_focus=true) in Start project and execute tests step

.github/workflows/e2e.yml
yurkovychv: All I need for starting test execution is to have the host up. As I understand we will have only one host even if we will have more packages

packages/platform-ui/cypress/plugins/index.js
yurkovychv: May I just remove the comments and leave this file as is? The project skeleton was generated by cypress. In order to avoid confusions in the future I'd leave this file

packages/platform-ui/cypress/support/index.js
yurkovychv: It actually was provided: https://github.com/percona-platform/saas-ui/blob/85cb2bd9b47e566e4d2017c0ad47838f4482b0ba/packages/platform-ui/cypress/support/index.js#L17

packages/platform-ui/cypress/support/commands.js
yurkovychv: Done thanks.

packages/platform-ui/cypress/support/index.js
yurkovychv: Done

yurkovychv: Added comments

packages/platform-ui/cypress/support/index.js
yurkovychv: Could you recheck once again? Is it ok now?

.github/workflows/e2e.yml
nicolalamacchia: Not sure I understand. Each package could potentially have a `start` script, which could launch its own local server. Anyway, let's keep this for now, but let's be mindful about this.

.github/workflows/e2e.yml
nicolalamacchia: Could you add this to that comment please? :) Also, please consider this is a workaround and it's not documented.

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,44 @@
...
+    let item = runnable;
+    const nameParts = [runnable.title];
+
+    while (item.parent) {
+      nameParts.unshift(item.parent.title);
+      item = item.parent;
+    }
+    if (runnable.hookName) {
+      nameParts.push(`${runnable.hookName} hook`);
+    }
```
nicolalamacchia, line 29: This is still very difficutl to understand, please try to convert it in a more readable code (using for example maps and reducers) or at least document each part in detail.

packages/platform-ui/cypress/integration/auth/signup.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    const password = 'Password123';
+    cy.get(auth.submitButton).should('be.visible').should('be.disabled');
+    cy.get(auth.emailField).clear().type(email);
+    cy.get(auth.emailValidation).should('have.text', '');
+    cy.get(auth.passwordField).type(password);
+    cy.get(auth.passwordValidation).should('have.text', '');
+    cy.get(auth.termsCheckbox).click({ force: true });
+    cy.get(auth.submitButton).should('be.visible').should('be.enabled').click();
+    cy.popUpContains('You have successfully created your credentials');
+    cy.url().should('contain', '/login')
```
nicolalamacchia, line 63:
```suggestion
    cy.url().should('contain', '/login');
```

packages/platform-ui/cypress/integration/auth/locators.js
```diff
@@ -0,0 +1,18 @@
...
+  emailValidation: 'qa:email-field-error-message',
+  passwordFieldLabel: 'qa:password-field-label',
+  passwordField: 'qa:password-password-input',
+  passwordValidation: 'qa:password-field-error-message',
+  submitButton: 'qa:login-submit-button',
+  signUpLink: 'qa:signup-action-button',
+  logoutButton: 'qa:logout-action-button',
+  termsCheckbox: 'qa:consent-checkbox-input',
+  termsText: 'qa:consent-field-label',
+  termsValidation: 'qa:consent-field-error-message'
```
nicolalamacchia, line 17:
```suggestion
  termsValidation: 'qa:consent-field-error-message',
```

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,44 @@
+import './commands';
+import addContext from 'mochawesome/addContext';
+import setup from 'cypress-cy-select';
+import {configure} from '@testing-library/cypress';
+
+// cypress-cy-select config
+const config = {
+  name: 'qa',
+  separator: ':'
```
nicolalamacchia, line 9:
```suggestion
  separator: ':',
```

packages/platform-ui/cypress/support/index.js
```diff
@@ -0,0 +1,44 @@
...
+};
+setup(config);
```
nicolalamacchia, line 11:
```suggestion

setup(config);
```

nicolalamacchia (changes requested):

packages/platform-ui/cypress/plugins/index.js
nicolalamacchia: If that's the case, you can leave it as it is. Please notice that you removed the `/// <reference types="cypress" />
` line in a recent commit

packages/platform-ui/cypress/plugins/index.js
```diff
@@ -0,0 +1,11 @@
+// This function is called when a project is opened or re-opened (e.g. due to
```
nicolalamacchia, line 1: I think `/// <reference types="cypress" />` it's missing

packages/platform-ui/cypress/plugins/index.js
yurkovychv: yeah, my bad

nicolalamacchia (changes requested):

packages/platform-ui/cypress/plugins/index.js
yurkovychv: just checked, It wasn't there

nicolalamacchia: Mmm, that's strange, I see it here: https://github.com/percona-platform/saas-ui/commit/124de59910b5245946975fd708250e7939d1e989#diff-4d8b973b2577a8f2d0b1e06f8e0d298a

yurkovychv: Oh yeah, I replied to this in the discussion above) thought this is another file

.github/workflows/e2e.yml
yurkovychv: sure, will keep in mind

packages/platform-ui/cypress/support/index.js
yurkovychv: updated the code, please check it

.github/workflows/e2e.yml
yurkovychv: Done

nicolalamacchia (approved):

packages/platform-ui/cypress/integration/auth/locators.js
```diff
@@ -0,0 +1,18 @@
+// to specify data-qa attribute selector add prefix 'qa:' to the locator
+
+export const authLocators = {
```
atymchuk, line 3: nit: It's generally a good idea to sort the keys in objects like this one

atymchuk (approved):

PR #30, opened 2020-09-28
SAAS-334 get email from refresh session API response

nicolalamacchia:
- [x] https://github.com/percona-platform/platform/pull/140
- [x] https://github.com/percona-platform/authed/pull/49

### DIFF
--- a/packages/platform-ui/src/store/auth/auth.reducer.ts
+++ b/packages/platform-ui/src/store/auth/auth.reducer.ts
@@ -1,5 +1,5 @@
 import { createAsyncAction, ActionType, getType } from 'typesafe-actions';
-import { AuthState } from '../types';
+import { AuthState } from 'store/types';
 
 const DEFAULT_STATE: AuthState = {
   authenticated: false,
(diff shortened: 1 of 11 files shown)


packages/platform-ui/src/store/auth/auth.reducer.ts
```diff
@@ -11,7 +11,7 @@ export const authRefreshAction = createAsyncAction(
   'LOGIN_REFRESH_REQUEST',
   'LOGIN_REFRESH_SUCCESS',
   'LOGIN_REFRESH_FAILURE',
-)<undefined, undefined, Error>();
+)<undefined, { email: string }, Error>();
```
tiagomotasantos, line 14: Maybe use interface for the action payload here

atymchuk (changes requested):

packages/platform-ui/src/store/auth/auth.reducer.ts
nicolalamacchia: Ok I did that, using `Pick<...>`

tiagomotasantos (approved):

tiagomotasantos (approved):

atymchuk (approved):

PR #32, opened 2020-10-01
SAAS-363 add redux saga

nicolalamacchia:


packages/platform-ui/src/index.tsx
```diff
@@ -29,9 +30,11 @@ ReactDOM.render(
   <React.StrictMode>
     <ThemeContext.Provider value={light}>
       <Provider store={store}>
-        <Router>
-          <Main />
-        </Router>
+        <ConnectedRouter history={history}>
+          <Router>
```
atymchuk, line 34: The router is not needed then. Pls see https://github.com/supasate/connected-react-router#step-3

packages/platform-ui/src/store/auth/auth.sagas.ts
```diff
@@ -0,0 +1,60 @@
...
+    yield call([toast, toast.error], action.payload.message);
+  } else {
+    yield call([toast, toast.error], Messages.api.signInFailed);
+    console.error(action.payload);
+  }
+}
+
+function* authLoginSuccess(action: ReturnType<typeof authLoginAction.success>): Generator<StrictEffect, void, never> {
+  yield call([toast, toast.success], `${Messages.api.signInSucceeded} ${action.payload.email}`);
+  yield call([history, history.replace], Routes.root);
```
atymchuk, line 50: This is not cosher, pls refer to https://github.com/supasate/connected-react-router/blob/master/FAQ.md#in-redux-saga

atymchuk (changes requested):

packages/platform-ui/src/index.tsx
nicolalamacchia: I made it work without it, but had to downgrade `history` @ 4.10.1 (see https://github.com/ReactTraining/history/issues/804 and https://github.com/supasate/connected-react-router/issues/312)

packages/platform-ui/src/store/auth/auth.sagas.ts
nicolalamacchia: Done

nicolalamacchia:
e2e tests must be fixed, @yurkovychv let's have a discussion about that if needed

yurkovychv (approved):

packages/platform-ui/package.json
```diff
@@ -38,23 +38,29 @@
...
     "emotion": "^10.0.27",
     "final-form": "^4.20.1",
     "google-protobuf": "^3.13.0",
     "grpc-web": "^1.2.1",
+    "history": "4.7.2",
     "react": "16.12.0",
     "react-dom": "16.12.0",
     "react-final-form": "^6.5.1",
     "react-redux": "^7.2.1",
+    "react-router": "^5.2.0",
```
atymchuk, line 57: Do we need it? It's an indirect dependency of _react-router-dom_ AFAIK, so we should probably leave it out.

atymchuk (changes requested):

yurkovychv (dismissed):

packages/platform-ui/package.json
nicolalamacchia: Right, it's a mistake, removed it

atymchuk (approved):

PR #33, opened 2020-10-05
SAAS-369 pass html attributes to inputs

atymchuk:
Note: this PR adds the ability to pass html attributes which are only relevant for a particular type of input. Bonus: ts autocomplete will reject invalid attributes. Added tests to cover this.

Added to utils: 
1/ a logger, so no more lint warnings :)
2/ a FormWrapper to DRY wrapping inputs in forms

Won't do: no test coverage for the console, which is difficult to test. Thorough manual testing only.
Todo: write a mock for the logger to free us from having to mock it everywhere.


packages/platform-core/src/components/Checkbox/CheckboxField.tsx
```diff
@@ -1,50 +1,50 @@
...
   return (
     <Field {...fieldConfig} type="checkbox" name={name} validate={validate}>
       {({ input, meta }: CheckboxFieldRenderProps) => (
         <div className={cx(styles.field, fieldClassName)} data-qa={`${name}-field-container`}>
-          {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
-          <label className={styles.wrapper}>
+          <label className={styles.wrapper} htmlFor={inputId}>
             <input
+              id={inputId}
+              {...inputProps}
```
tiagomotasantos, line 46: Should we place {...inputProps} below {...input}? Because the way it currently is I think we won't be able to override the props react final form passes by default. For example it passes onChange by default in the input so it will always override our own even if we pass it as part of inputProps.

.gitignore
```diff
@@ -4,5 +4,13 @@ bin
...
 
+# IDE
+.vscode
+.idea
+
 # misc
 .DS_Store
+
+# IDEs
+.vscode
```
tiagomotasantos, line 15: These two lines are duplicated above.

tiagomotasantos (changes requested):
Well done with the utility test functions (FormWrapper and dataQa) introduced in this PR as well as the other great cleanups introduced 🎉👏

packages/platform-core/src/components/Checkbox/CheckboxField.tsx
atymchuk: I was struggling myself with it and came to a conclusion that this is not the way react-final-form is designed. Normally, if you need to hook to every change event on a particular field, you would grab the values from the render props and do what you need to do with any of the form values.

One major observation I have to make here is: we are paying a certain price of wrapping components in `Field`. As always, there are benefits and downsides. The great benefit is the ease of use - it is a pleasure to work with it compared to adapters we have in PMM. The downside is: by hiding the boilerplate it offers much less, it's opinionated. But that was exactly the purpose.

I would say, it is not universal on purpose and may only cover 90% of our needs. 

Please feel free to suggest a better design (or refactor/upgrade maybe), I'll be happy to review and I do acknowledge this one is not ideal.

.gitignore
atymchuk: Fixed

packages/platform-core/src/components/Checkbox/CheckboxField.tsx
tiagomotasantos: Ok, I understand that using onChange for example is not the final form way of doing it, I just don't understand what's the downside of changing the order of {...input} and {...inputProps}. Then we give the flexibility to the user of our library to do whatever he wants. Unless you are saying that we specifically don't want to give that flexibility, then I understand :D

atymchuk: That was exactly it: we don't want the user to override the native `onChange`, since it breaks the component. At least, this is what I tried and failed.

packages/platform-core/src/shared/logger/logger.ts
```diff
@@ -0,0 +1,27 @@
...
+export const setLogLevel = (level: LOG_LEVELS) => {
+  CONFIG_LOG_LEVEL = level;
+};
+
+const createLogMethod = (
+  loggerFunc: (...attrs: Array<any>) => void,
+  level: LOG_LEVELS,
+) => (...attrs: Array<any>) => {
+  if (level >= CONFIG_LOG_LEVEL) {
+    console.group(level, truncate(25)(attrs[0]));
```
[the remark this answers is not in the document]
atymchuk: > it's not clear to me, why just the first attribute? does it have a special meaning?

Yes, it's sort of a `headline` to enable the user to quickly find the one he is looking for without forcing them to read the full log.

> what's the benefit of having a console group here? it'd be almost the same also when passing several attributes to the console function

The `console.group` nicely indents the text, so it makes the console output easier to navigate. 

> I don't really see the benefit, it just looks more verbose to me

I'll paste an example to visualize the benefit. There used to be a blue icon for `console.info`, and I no longer see it :(
I also updated the code, so it should make more sense now.

Screenshot 2020-10-22 at 17 55 46

packages/platform-core/src/components/Checkbox/CheckboxField.tsx
atymchuk: We will try to weight the pros and cons in a live-coding session tomorrow before committing.

packages/platform-core/src/shared/logger/logger.ts
```diff
@@ -0,0 +1,34 @@
...
+  loggerFunc: (...attrs: Array<any>) => void,
+  level: LOG_LEVELS,
+) => (first: any, ...rest: Array<any>) => {
+  if (level >= CONFIG_LOG_LEVEL) {
+    const key = LOG_LEVEL_KEYS[level];
+
+    console.group(`[${key}]`, truncate(50)(first));
+    loggerFunc(first);
+    rest.forEach(para => {
+      loggerFunc(para);
```
nicolalamacchia, line 23: why not `param`? :D

atymchuk: it's a typo!

packages/platform-core/src/shared/logger/logger.ts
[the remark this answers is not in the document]
nicolalamacchia: Ok, now it makes more sense to have it like this :) this means one param per line though, we should document this somehow I think

packages/platform-core/src/components/Checkbox/CheckboxField.tsx
atymchuk: @nicolalamacchia The experiment showed that we can reverse the order so that {...inputProps} will win over {...input}. So, it's safe now to hook on the input's `change, blur, focus` events.

packages/platform-core/src/components/Checkbox/CheckboxField.tsx
```diff
@@ -37,14 +37,14 @@ export const CheckboxField: FC<CheckboxProps> = React.memo(({
   ]);
 
   return (
-    <Field {...fieldConfig} type="checkbox" name={name} validate={validate}>
+    <Field<boolean> {...fieldConfig} type="checkbox" name={name} validate={validate}>
```
atymchuk, line 40: @nicolalamacchia pls note, <Field> has a generic :)

packages/platform-core/src/shared/logger/logger.ts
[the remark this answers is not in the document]
atymchuk: Exactly, one param per line is a feature lots of dev are looking for in a logger :)

tiagomotasantos (approved):

nicolalamacchia (approved):

PR #40, opened 2020-10-12
SAAS-362 UI components tests

nicolalamacchia:


packages/platform-core/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { FC } from 'react';
+import { mount } from 'enzyme';
+import { CheckboxField } from './CheckboxField';
+import { Form, Field } from 'react-final-form';
+
+jest.spyOn(console, 'error').mockImplementation(() => {});
```
atymchuk, line 6: There is way to avoid putting this line in every test. 
1/ export console to /shared
2/ put a mock to its __mock__ folder
3/ use it without ^ that line :)

packages/platform-core/src/shared/validators/containsNumbers.test.ts
```diff
@@ -0,0 +1,9 @@
+import { containsNumber } from './containsNumber';
+
+describe('validators :: containsNumber', () => {
+  it('returns an error if there are no numbers characters in the passed string', () => {
+    expect(containsNumber('123')).toEqual(undefined);
```
atymchuk, line 5: We need to test edge cases by passing undefined, {}, [], '', etc
The validator may fail, btw. Then we should make it robust :)

atymchuk (changes requested):

packages/platform-core/src/shared/validators/email.test.ts
```diff
@@ -0,0 +1,25 @@
+import { email } from './email';
+
+describe('validators :: email', () => {
+  // NOTE (nicolalamacchia): some of these tests were taken from Chromium's source code
```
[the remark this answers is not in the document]
nicolalamacchia: None taken, actually I agree with you, done!

packages/platform-core/src/components/Checkbox/Checkbox.test.tsx
nicolalamacchia: we will introduce a better logging system, deprecating the use of `console`

packages/platform-core/src/shared/validators/containsNumbers.test.ts
atymchuk: Ok, let's give it another iteration. Looks acceptable at this stage. :)

nicolalamacchia: We have to harden our validators first, and subsequently improve test coverage

packages/platform-ui/src/components/Authenticated/Authenticated.tsx
```diff
@@ -24,7 +25,7 @@ export const Authenticated: FC = () => {
         <p>{Messages.authenticatedFirstLine}</p>
         <p>
           <em>
-            <b>{email}</b>
+            <b data-qa="user-email">{email}</b>
```
atymchuk, line 28: 👍

atymchuk (approved):

PR #41, opened 2020-10-15
SAAS-398 add eslint config

nicolalamacchia:
Add the same eslint configuration PMM uses
Fix linting problems according to the new eslint configuration
Add `lint` and `lint:fix` yarn/npm scripts
⚠️ Grafana's eslint plugin has been disabled in platform-ui, since it causes an issue in the build process. This needs further investigation.


packages/platform-core/src/components/Checkbox/CheckboxField.tsx
```diff
@@ -1,5 +1,7 @@
 import React, { FC, useMemo, ReactNode } from 'react';
-import { Field, FieldInputProps, FieldMetaState, UseFieldConfig } from 'react-final-form';
+import {
+  Field, FieldInputProps, FieldMetaState, UseFieldConfig
```
tiagomotasantos, line 3: With `only-multiline` a comma should be required here or am I wrong?

nicolalamacchia: only if they end up being one on each line :)

tiagomotasantos: Great :)

nicolalamacchia: But you made me notice that there's a but, it should be `always-multiline` :D

packages/platform-ui/.eslintrc
```diff
@@ -0,0 +1,92 @@
...
+      { "blankLine": "always", "prev": "for", "next": "*" },
+      { "blankLine": "always", "prev": ["const", "let", "var"], "next": "*" },
+      { "blankLine": "any", "prev": ["const", "let", "var"], "next": ["const", "let", "var"] }
+    ],
+    // "no-inline-styles/no-inline-styles": [2],
+    // ENABLE LATER
+    "max-classes-per-file": ["off"],
+    "no-shadow": ["off"],
+    "react/no-array-index-key": ["off"],
+    "@typescript-eslint/ban-ts-ignore": ["off"],
```
atymchuk, line 88: Consider applying the same rules an in platform-core (in case they are applied there).

atymchuk (changes requested):

packages/platform-core/.eslintrc
```diff
@@ -0,0 +1,91 @@
...
+    // We use a lot of snake case variables in code interacting with API
+    "camelcase": ["off"],
+    // It's more convinient to declare big helpers functions below the main method
+    "no-use-before-define": ["off"],
+    // Don't need these
+    "jsx-a11y": ["off"],
+    "jsx-a11y/anchor-is-valid": ["off"],
+    "jsx-a11y/click-events-have-key-events": ["off"],
+    "jsx-a11y/no-static-element-interactions": ["off"],
+    "jsx-a11y/no-noninteractive-element-interactions": ["off"],
```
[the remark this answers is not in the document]
nicolalamacchia: This is the same as PMM, should we switch it on there too?

packages/platform-core/.eslintrc
```diff
@@ -0,0 +1,91 @@
...
+      { "blankLine": "any", "prev": ["const", "let", "var"], "next": ["const", "let", "var"] }
+    ],
+    // "no-inline-styles/no-inline-styles": [2],
+    // ENABLE LATER
+    "max-classes-per-file": ["off"],
+    "no-shadow": ["off"],
+    "react/no-array-index-key": ["off"],
+    "@typescript-eslint/ban-ts-ignore": ["off"],
+    // enable it later, it happens because null coalesion operator
+    "no-unused-expressions": ["off"]
```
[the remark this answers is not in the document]
nicolalamacchia: No, I just copied the rules from PMM :)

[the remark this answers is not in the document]
atymchuk: In fact, we should. To get more value out of this PR.

packages/platform-core/.eslintrc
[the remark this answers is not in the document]
atymchuk: Not in PMM, we don't have to keep it in perfect sync.

atymchuk (approved):

PR #42, opened 2020-10-20
SAAS-405 dropdown component

nicolalamacchia:


packages/platform-core/src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+    unmountComponentAtNode(container);
+    container.remove();
+  });
+
+  test('should render the toggle', () => {
+    act(() => {
+      render(<Dropdown toggle={Toggle}><a href="/">root</a><a href="/test">test</a></Dropdown>, container);
+    });
+
+    expect(container.querySelector('[data-qa="dropdown-toggle"]')).not.toBe(null);
```
atymchuk, line 31: nit: Consider using `dataQa` fn for brevity sake (here and downwards)

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+    document.addEventListener('mousedown', handleDocumentClick);
+
+    return () => {
+      document.removeEventListener('mousedown', handleDocumentClick);
+    };
+  }, []);
+
+  return (
+    <>
+      <Toggle ref={toggleRef} onClick={handleDropdownClick} data-qa="dropdown-toggle" />
```
atymchuk, line 72:
```suggestion
      <Toggle ref={toggleRef} onClick={handleDropdownClick} data-qa="dropdown-menu-toggle" />
```
Since may have other dropdown components, consider make it more specific

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+      <div
+        ref={popperRef}
+        style={popperStyles.popper}
+        {...popperAttributes.popper}
+        data-qa="dropdown-container"
```
atymchuk, line 78:
```suggestion
        data-qa="dropdown-menu-container"
```
To stay consistent with the above ^

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+      >
+        {visible ? (
+          <div
+            className={cx(styles.dropdownMenu, className)}
+            style={popperStyles.offset}
+            data-qa="dropdown-menu"
```
atymchuk, line 84:
```suggestion
            data-qa="dropdown-menu-menu"
```
For consistency again

packages/platform-core/src/components/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './StepProgress';
 export * from './TextInput';
 export * from './TextareaInput';
 export * from './NumberInput';
+export * from './Dropdown';
```
atymchuk, line 8: nit: let's sort them ab

atymchuk (changes requested):

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const toggleRef = useRef<HTMLElement>(null);
+  const popperRef = useRef<HTMLDivElement>(null);
+
+  const { styles: popperStyles, attributes: popperAttributes } = usePopper(
+    toggleRef.current,
+    popperRef.current,
+    popperConfig,
+  );
+
+  const handleDocumentClick = (event: MouseEvent) => {
```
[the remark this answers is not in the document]
nicolalamacchia: Should they really be wrapped? This function specifically is used only internally to the functional component, while the other is setting a state. I think performance won't benefit, or can even get worse

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (
+      event.target instanceof Node &&
```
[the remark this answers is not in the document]
nicolalamacchia: That's needed mostly for type hinting on the arguments of the following `contains`. I guess we could just cast them to type `Node` if we want to get rid of that check

packages/platform-core/src/components/Dropdown/Dropdown.test.tsx
nicolalamacchia: done

packages/platform-core/src/components/index.ts
nicolalamacchia: done

packages/platform-core/src/components/Dropdown/Dropdown.tsx
nicolalamacchia: done

packages/platform-core/src/components/Dropdown/Dropdown.tsx
nicolalamacchia: done

packages/platform-core/src/components/Dropdown/Dropdown.tsx
nicolalamacchia: done

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+      enabled: true,
+      options: {
+        offset: [0, 10],
+      },
+    },
+  ],
+};
+
+interface DropdownProps {
+  toggle: React.ForwardRefExoticComponent<any>;
```
tiagomotasantos, line 22: Does it really need to be any here? Or some type like ReactNode or HTMLElement would work?

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+    <>
+      <Toggle ref={toggleRef} onClick={handleDropdownClick} data-qa="dropdown-menu-toggle" />
+
+      <div
+        ref={popperRef}
+        style={popperStyles.popper}
+        {...popperAttributes.popper}
+        data-qa="dropdown-menu-container"
+      >
+        {visible ? (
```
tiagomotasantos, line 80: If you replace "?" with "&&" you can remove the unused else part of the condition here.

tiagomotasantos (changes requested):

packages/platform-core/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
atymchuk: Agree, casting to `Node` seems to solve it.

packages/platform-core/src/components/Dropdown/Dropdown.tsx
nicolalamacchia: I agree with you, but soon after I joined we had a discussion and decided to prefer the ternary operator to be more consistent and flexible (since you can't always use micro-branching)

packages/platform-core/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
atymchuk: I agree, this conversation would have a more solid base in case we approach it with: `I profiled it and here is what I found`. But that's for the asking party, which is something I didn't plan on doing right now :) 
So yes, let's leave it as is.

packages/platform-core/src/components/Dropdown/Dropdown.tsx
nicolalamacchia: They would, partially. The passed component needs a forwarded `ref`, otherwise we cannot refer to it in the dropdown

packages/platform-core/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
nicolalamacchia: done

packages/platform-core/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
nicolalamacchia: :D ok

packages/platform-core/src/components/Dropdown/Dropdown.tsx
tiagomotasantos: What that has to do with the any? (sorry just trying to understand if something like this would work React.ForwardRefExoticComponent<ReactNode> or React.ForwardRefExoticComponent<HTMLElement>)

nicolalamacchia: Ohhh, sorry, I misread your first comment! I updated the type there :D

tiagomotasantos: Oh okay 😁 got it, great!

tiagomotasantos (approved):

tiagomotasantos (approved):

atymchuk (approved):

atymchuk:
@nicolalamacchia It needs one more test to cover the keyboard nav :)

nicolalamacchia:
@atymchuk it does, but this has already gone way too far from the original estimation. We started with an investigation on a library and instead ended up developing our own dropdown. I suggest to deal with this separately.

tiagomotasantos (dismissed):

atymchuk:
> @atymchuk it does, but this has already gone way too far from the original estimation. We started with an investigation on a library and instead ended up developing our own dropdown. I suggest to deal with this separately.

1SP is arguably a good estimate for such a library component. It should be much higher given the inherent complexity of it.

nicolalamacchia:
> 1SP is arguably a good estimate for such a library component. It should be much higher given the inherent complexity of it.

For an homemade component I agree, it was a bad estimation, but again the original idea was to just research for a third party one and include it in our library :)

packages/platform-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,135 @@
...
+    document.addEventListener('keyup', handleKeyupClick);
+    document.addEventListener('keydown', handleKeydownClick);
+
+    return () => {
+      document.removeEventListener('mousedown', handleDocumentClick);
+      document.removeEventListener('keyup', handleKeyupClick);
+      document.removeEventListener('keydown', handleKeydownClick);
+      // UX: reset the index when we tear down the children
+      setActiveIndex(-1);
+    };
```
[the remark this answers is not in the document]
atymchuk: yes, I just didn't have enough stamina to do that ) 
Done!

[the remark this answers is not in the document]
atymchuk: > we're not listening for enter or other keyboard inputs to simulate the click on that entry.
What do you mean?

atymchuk (approved):

packages/platform-core/src/components/Dropdown/Dropdown.tsx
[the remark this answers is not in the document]
nicolalamacchia: well, I can scroll using arrow keys, but I cannot action on menu items. What's the point of having key navigation if the only thing I can do is to highlight menu items, but I still have to click on those with my mouse? :)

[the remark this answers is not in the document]
atymchuk: Wow, that's a good point! we'll need to get that implemented for sure :)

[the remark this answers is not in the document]
atymchuk: Done as well!

atymchuk (dismissed):

atymchuk (approved):

PR #44, opened 2020-10-29
SAAS-393 execute e2e at minikube

yurkovychv:
updated `ci.yml` workflow and added e2e execution job after Building Docker Image. e2e.yml workflow executes tests with `latest` tags nightly


.github/workflows/ci.yml
```diff
@@ -110,3 +110,10 @@ jobs:
 
       - name: Push Docker image
         run: make docker-push
+
+      - name: Trigger E2E execution with tag "${{ env.DOCKER_TAG }}"
+        uses: percona-platform/repository-dispatch@v1
```
AlekSi, line 115: Why not use https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_dispatch directly?

AlekSi: Err, that's confusing :) What I really wanted to ask is why use repository_dispatch instead of workflow_dispatch?

yurkovychv: There is no particular reason. Will try with workflow_dispatch

AlekSi (changes requested):
We can't just clone infra repository there and copy&paste all setup instructions there – it would be very hard to keep them in sync, especially once we copy&paste in other repos too.

We need find solution that will run tests in the context of infra repo or separate cicd repo. We talked about workflow_dispatch and repository_dispatch – was there any follow up?

.github/workflows/ci.yml
AlekSi: Why that was abandoned?

yurkovychv:
I need a spike for that, I was testing it with workflow_dispatch. But it does not affect PR checks (if tests failed a PR does not indicate that, and it is not clear which exact PR triggered that workflow). It just says `Manually run by percona-robot`. The best way would be to have 1 workflow in `cicd` repo. But again we need to find a way to detect which PR triggered it (without reading logs) and indicate it in the original PR.

.github/workflows/ci.yml
yurkovychv: I replied [here](https://github.com/percona-platform/saas-ui/pull/44#issuecomment-720983995)

.github/workflows/e2e.yml
```diff
@@ -29,51 +93,42 @@ jobs:
...
-          start: yarn start
-          wait-on: 'http://localhost:3000'
-          wait-on-timeout: 120
+          working-directory: saas-ui/packages/platform-ui
+          command: yarn cy:run
 
       - name: Generating the report
-        if: success() || failure()
-        working-directory: packages/platform-ui
+        if: ${{ always() }}
```
nicolalamacchia, line 110: Why is this needed?

.github/workflows/e2e.yml
```diff
@@ -29,51 +93,42 @@ jobs:
         id: cache
         with:
           path: |
-            node_modules
-            */*/node_modules
+            saas-ui/node_modules
+            */*/*/node_modules
```
nicolalamacchia, line 97: would `**/nodemodules` work here?

AlekSi (approved):

.github/workflows/e2e.yml
AlekSi: It can also be `canceled`, and it is different from being failed

yurkovychv: Yes, we need this step to be executed anyway

.github/workflows/e2e.yml
yurkovychv: not sure, should I try It now? Anyway, this workflow will be moved to another repo and updated soon.

nicolalamacchia (approved):

PR #51, opened 2020-11-11
SAAS-443 add security headers to nginx

atymchuk:


AlekSi:
We should add those headers to Traefik since API responses will be affected too

atymchuk:
Will be done in traefik's config

PR #57, opened 2020-11-18
PMM-6925 table toolbar

nicolalamacchia:


packages/platform-core/src/components/TableToolbar/TableToolbar.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { getStyles } from './TableToolbar.styles';
+import { Messages } from './TableToolbar.messages';
+import { TableToolbarProps } from './types';
+
+export const TableToolbar: FC<TableToolbarProps> = ({ actions, selectedItems }) => {
+  const styles = useMemo(() => getStyles(), []);
+
+  return (
+    <div className={styles.wrapper}>
+      {actions.map(action => {
```
tiagomotasantos, line 13: Why not destructure the action here? 🤔

packages/platform-core/src/components/TableToolbar/types.ts
```diff
@@ -0,0 +1,20 @@
+import { AvailableIcons } from 'components/Icon';
+
+export interface Action {
+  callback: () => {};
```
tiagomotasantos, line 4: Does the callback return an empty object or is void?

tiagomotasantos (dismissed):

packages/platform-core/src/components/TableToolbar/TableToolbar.tsx
nicolalamacchia: Oops, right, done

packages/platform-core/src/components/TableToolbar/types.ts
nicolalamacchia: it's `void`, updated it :)

tiagomotasantos (approved):

packages/platform-core/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+import {
+  TableToolbarAdd,
+  TableToolbarDelete,
+  TableToolbarEnable,
+  TableToolbarDisable,
+} from '../../shared/icons';
+
+// TODO: improve this mapping
+const icons: Icons  = {
+  toolbarAdd: TableToolbarAdd,
```
atymchuk, line 13: If an icon can be used elsewhere, then it's better to give it a name that does not contain the domain (ie. `toolbar` and `table`). 
However, since the `Add` icons can be of different shapes (ex: circled, square, etc) we should hint the dev about the shape. => `AddSquare` or `AddCircle`.
Grafana got it right, good naming approach here https://developers.grafana.com/ui/latest/index.html?path=/story/docs-overview-icon--icons-overview

packages/platform-core/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+  toolbarDelete: TableToolbarDelete,
+  toolbarEnable: TableToolbarEnable,
+  toolbarDisable: TableToolbarDisable,
+};
+
+interface IconProps {
+  name: AvailableIcons;
+}
+
+export const Icon: FC<IconProps> = ({ name }) => {
```
atymchuk, line 23: What about passing basic props, for example to change the icon size? Seems relevant

packages/platform-core/src/components/TableToolbar/TableToolbar.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { FC, useMemo } from 'react';
+import { error } from '../../shared/logger';
+import { TableToolbarButton } from './TableToolbarButton';
+import { getStyles } from './TableToolbar.styles';
+import { Messages } from './TableToolbar.messages';
+import { TableToolbarProps } from './types';
+
+export const TableToolbar: FC<TableToolbarProps> = ({ actions, selectedItems }) => {
+  const styles = useMemo(() => getStyles(), []);
```
atymchuk, line 9: This seems to be contrary to the recommendation (https://github.com/grafana/grafana/blob/master/contribute/style-guides/themes.md#usestyles-hook). useStyles is memoized. And it's probably not compatible with the theme switch.

packages/platform-core/src/components/TableToolbar/TableToolbarButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { TableToolbarButtonProps } from './types';
+
+export const TableToolbarButton: FC<TableToolbarButtonProps> = ({
+  className,
+  icon,
+  label,
+  ...props
+}) => {
+  const theme = useTheme();
+  const styles = useMemo(() => getStyles(theme), [theme]);
```
atymchuk, line 15: should be useStyles; both useTheme & useMemo are not necessary.

packages/platform-core/src/components/TableToolbar/types.ts
```diff
@@ -0,0 +1,20 @@
+import { AvailableIcons } from 'components/Icon';
+
+export interface Action {
+  callback: () => void;
+  icon: AvailableIcons;
+  label: string;
+  minItems?: number;
```
atymchuk, line 7: Sorry, how the min/max are used?

atymchuk (changes requested):

packages/platform-core/src/components/TableToolbar/types.ts
nicolalamacchia: it's the minimum/maximum number of items to be selected for the action to be enabled

packages/platform-core/src/components/Icon/Icon.tsx
nicolalamacchia: I'd prefer to keep this outside of the scope of the ticket

packages/platform-core/src/components/TableToolbar/TableToolbar.tsx
nicolalamacchia: oh, I forgot about that! Fixed

packages/platform-core/src/components/TableToolbar/TableToolbarButton.tsx
nicolalamacchia: Done

packages/platform-core/src/components/TableToolbar/types.ts
atymchuk: In this case we don't need it. There is simply no use case for anything other than `check if [...selected].length > 1`.
Let's keep it simple, which should reduce the amount of code and the testing effort.

nicolalamacchia: there might be actions needing just one element, while other needing > 1, or did I get this wrong? 😅 In that case, I'd agree with you

packages/platform-core/src/components/Icon/Icon.tsx
nicolalamacchia: Good point, done!

packages/platform-core/src/components/TableToolbar/types.ts
atymchuk: By design, actions applicable for just one element would go to the `Actions` column. This is place for either mass actions, i.e. which need >= 2 records to be fired, or what we call `one-off` actions (like `Add a record`), which do not depend on the the number of records selected. 
In other words, I don't know of a use case, where `min` would be 1 or 3. It's always `two or more records` for the mass actions.

packages/platform-core/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,27 @@
...
+  PlusSquare,
+  SelectedSquare,
+  UnselectedSquare,
+} from '../../shared/icons';
+
+// TODO: improve this mapping
+const icons: Icons  = {
+  plusSquare: PlusSquare,
+  minusSquare: MinusSquare,
+  selectedSquare: SelectedSquare,
```
atymchuk, line 15: From the goal standpoint, this should be `Enable`, the one below would be `Disable`. And we need them to have meaningful names :)

packages/platform-core/src/components/TableToolbar/TableToolbar.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <div className={styles.wrapper}>
+      {actions.map(({
+          callback,
+          icon,
+          minItems = 0,
+          maxItems = Infinity,
+          label,
+        }) => {
+
+        if (minItems < 0 || maxItems < 0) {
```
atymchuk, line 22: I'm sure we don't have this in the scope. We enable/disable the mass action icons when the number of records selected is `>1`. So the user wouldn't be able to fire a mass action otherwise, which would make the error messages redundant.

atymchuk (changes requested):

packages/platform-core/src/components/Icon/Icon.tsx
nicolalamacchia: Done

packages/platform-core/src/components/TableToolbar/types.ts
nicolalamacchia: got it, done

packages/platform-core/src/components/TableToolbar/TableToolbar.tsx
atymchuk: @nicolalamacchia so, what do you think?

nicolalamacchia: I already did remove the error : )

tiagomotasantos (approved):

atymchuk (approved):

PR #60, opened 2020-11-27
SAAS-403 light/dark theme switch

atymchuk:
* light/dark theme switch
* localStorage theme persistence
* redux-based storage for the theme, saga + tests
* refactored markup for Menu & Authenticated components
* replace `img` with svg logos
* /ui page scaffold to demo components 
* /ui demo the TextInputField component
* much simpler builds for the core
* fixed front-page theme incompatibility
* fixed unit tests
* the bundle size went down from 6MiB to only 1.5MiB !
Screenshot 2020-12-02 at 00 25 10


packages/platform-ui/src/components/Root/Main.tsx
```diff
@@ -1,46 +1,35 @@
...
           </PrivateRoute>
-          <PublicRoute path={Routes.login}>
+          <PublicRoute exact path={Routes.login}>
             <LoginPage />
           </PublicRoute>
-          <PublicRoute path={Routes.signup}>
+          <PublicRoute exact path={Routes.signup}>
             <SignupPage />
           </PublicRoute>
-          <Route path="*">
```
tiagomotasantos, line 35: Do we want to remove the "not found" route?

packages/platform-ui/src/store/theme/theme.reducer.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type ThemeActions = (
+  ActionType<typeof themeChangeRequestAction> |
+  ActionType<typeof themeChangeSuccessAction>
+);
+
+export function themeReducer(state: ThemeState = DEFAULT_STATE, action: ThemeActions): ThemeState {
+  switch (action.type) {
+    case getType(themeChangeSuccessAction): {
+      if (state.isDark && action.payload.isDark) {
```
tiagomotasantos, line 23: Why not use `||` here with the condition form below?

tiagomotasantos (approved):
Good job 💪

packages/platform-ui/src/store/theme/theme.reducer.ts
atymchuk: Trying to avoid long boolean expressions in favour of readability :)

packages/platform-ui/src/components/Root/Main.tsx
atymchuk: Actually yes, because it's not ready yet, i.e. looks ugly )

nicolalamacchia (approved):

yurkovychv (approved):

PR #66, opened 2021-01-13
SAAS-385 UI 404

fabio-silva:
**What this PR does / why we need it:** Adds a 404 "Not Found" page when an authenticated user tries to reach an inexistent page (not authenticated users keep being redirected to login). This page follows the theme rules, being either light or dark.

**Which issue(s) this PR fixes:** https://jira.percona.com/browse/SAAS-385


packages/platform-ui/src/assets/404-dark.svg
```diff
@@ -0,0 +1,56 @@
+<svg width="532" height="442" viewBox="0 0 532 442" fill="none" xmlns="http://www.w3.org/2000/svg">
```
nicolalamacchia, line 1: Please, format this file and the other SVG

packages/platform-ui/src/components/Root/Main.tsx
```diff
@@ -28,7 +29,9 @@ export const Main: FC = () => {
           <PublicRoute path={Routes.ui}>
             <UIDemo />
           </PublicRoute>
-          <Redirect to={Routes.login} />
+          {
+            authenticated ? <NotFound /> : <Redirect to={Routes.login} />
+          }
```
nicolalamacchia, line 34: Please rely on our eslint rules and our formatting rules, formatting should be consistent

nicolalamacchia, line 34:
```suggestion
          {authenticated ? <NotFound /> : <Redirect to={Routes.login} />}
```

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
```diff
@@ -0,0 +1,30 @@
+import { css } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+
+export const getStyles = (theme: GrafanaTheme) => {
+  const {
+    colors: { dashboardBg },
+    spacing: { lg: spacingLg },
+  } = theme;
```
nicolalamacchia, line 8: check formatting here as well

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  return {
+    contentWrapper: css`
+      display: flex;
+      flex-direction: column;
+      min-height: 100vh;
+      background-color: ${dashboardBg};
+      align-items: center;
+      padding: 2em; // TODO set this on core and change everywhere, as it's used several times
```
nicolalamacchia, line 17: the comment should go on its own line

packages/platform-ui/src/pages/NotFound/NotFound.tsx
```diff
@@ -1,10 +1,26 @@
...
+  const { isDark } = useSelector(getCurrentTheme);
+  const styles = useStyles(getStyles);
+  const dataQaImgAttr = isDark ? 'dark' : 'light';
+  const logo = isDark ? darkLogo : lightLogo;
+
   return (
-    <div>
-      <h2>{Messages.pageNotFound}</h2>
+    <div className={styles.contentWrapper}>
+      <img data-qa={`404-image-${dataQaImgAttr}`} className={styles.logo} alt="404" src={logo} />
```
nicolalamacchia, line 20: I suggest you use a static `data-qa` attribute, to make it more predictable, we are not going to check if the theme is dark or light here.

packages/platform-ui/src/styles/global.css
```diff
@@ -37,6 +37,11 @@ ul {
   padding: 0;
 }
 
+a {
+  text-decoration: none;
+  color: inherit;
+}
+
```
nicolalamacchia, line 44: do we really need this? if so, why?

nicolalamacchia (changes requested):

packages/platform-ui/src/assets/404-dark.svg
fabio-silva: 👍

packages/platform-ui/src/components/Root/Main.tsx
fabio-silva: You're right! But the linter doesn't complain about this one. It seems we're missing the `react/jsx-curly-newline` rule. Should I open a PR adding this?

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
fabio-silva: Got it 👍

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
fabio-silva: 👍

packages/platform-ui/src/pages/NotFound/NotFound.tsx
fabio-silva: Ok, if we're not checking the theme, I guess it makes sense to remove it

packages/platform-ui/src/styles/global.css
fabio-silva: So, without this rule, this is the result:

Screenshot 2021-01-13 at 22 21 17

I thought about setting it globally because usually these kind of anchor styles apply everywhere. I can tell it's being applied locally [here](https://github.com/percona-platform/saas-ui/blob/main/packages/platform-ui/src/pages/Login/Login.styles.ts#L46).
Should I keep it local?

nicolalamacchia: Yes, please avoid global style unless strictly needed (to reset browsers' CSS or set style for third party components that require that). That style will be refactored when a `Button` component will be created.

packages/platform-ui/src/components/Root/Main.tsx
nicolalamacchia: No, try to keep PRs tied to tickets as much as possible. If needed, create a separate ticket. In this case you could do it in this PR, since the change is minimal :)

fabio-silva: Sure, agreed :))

packages/platform-ui/src/styles/global.css
fabio-silva: Ok, I guess that Button could wrap those styles afterwards 👍

packages/platform-ui/src/pages/NotFound/NotFound.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    container.remove();
+    jest.clearAllMocks();
+  });
+
+  test('has the 404 image', () => {
+    act(() => {
+      render(<TestContainer><NotFound /></TestContainer>, container);
+    });
+
+    expect(container.querySelector('[data-qa="404-image"]')).not.toBeNull();
```
nicolalamacchia, line 34: you can use the utility `dataQa` here from `platform-core`

fabio-silva: Changed here and in Main.tsx 👍

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
```diff
@@ -0,0 +1,41 @@
+import { css } from 'emotion';
+import { GrafanaTheme } from '@grafana/data';
+
+export const getStyles = (theme: GrafanaTheme) => {
```
tiagomotasantos, line 4: theme can be destructured here

tiagomotasantos (changes requested):

packages/platform-ui/src/components/Root/Main.tsx
```diff
@@ -28,7 +29,7 @@ export const Main: FC = () => {
           <PublicRoute path={Routes.ui}>
             <UIDemo />
           </PublicRoute>
-          <Redirect to={Routes.login} />
+          {authenticated ? <NotFound /> : <Redirect to={Routes.login} />}
```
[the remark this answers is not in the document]
fabio-silva: Nice one! Seems reasonable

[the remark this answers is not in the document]
nicolalamacchia: That's actually the right way of doing that :D good catch @tiagomotasantos

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+  const {
+    colors: {
+      dashboardBg,
```
atymchuk, line 7: nit: I don't think we should run the cost of too much wrapping here - it takes too long to read it.
It's totally fine not to destructure `dashboardBg` and `spacingLg` and use it in the code as `colors.dashboardBg` etc. Prefer explicit over guessing where the `dashboardBg` comes from: colors or palette.

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  return {
+    contentWrapper: css`
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+      min-height: 100vh;
+      background-color: ${dashboardBg};
+      align-items: center;
+      // TODO set this on core and change everywhere, as it's used several times
```
atymchuk, line 22: What does this comment refer to? It doesn't say how, so not enough context to fix it for someone later. As a result, this will pile up and never get fixed. The core is in the same repo, so why not fix it right away?

packages/platform-ui/src/pages/NotFound/NotFound.test.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import { render, unmountComponentAtNode } from 'react-dom';
```
atymchuk, line 2: No need to be using react-dom directly, since it's pretty low-level and requires more dance.
Pls check out best practices for [react-dom/test-utils](https://testing-library.com/docs/react-testing-library/example-intro/#full-example) and a good example [here](https://github.com/percona-platform/saas-ui/blob/main/packages/platform-ui/src/components/MenuBar/MenuBar.test.tsx).

Benefits:
1/ no need to create a container with document.createElement and to destroy it
2/ much less code, more readable
3/ fewer imports

atymchuk (changes requested):

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
fabio-silva: I see your point. Any recommendation regarding the level at which this padding should be defined? Because [here](https://github.com/percona-platform/saas-ui/blob/main/packages/platform-core/src/theming/default.ts) I see paddings defined both in `spacing` and as standalone properties like `panelPadding`. Or should paddings form a category of their own inside `spacing`, similar to what's done by [style-dictionary](https://github.com/amzn/style-dictionary/blob/master/examples/complete/properties/size/padding.json) ?

packages/platform-ui/src/pages/NotFound/NotFound.test.tsx
fabio-silva: You're right, thanks for the tip!

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
fabio-silva: Thanks!

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
atymchuk: Good question. It depends on where it belongs in. For example, if this padding is part of the form elements and it provides padding around those elements, it may live in `spacing.formElementPadding` (as an example, not a true thing), but this one is part of `contentWrapper`, so I'm not event sure we would ever have the same paddings on all content wrappers, which we may have thousands of in one app.

Maybe, it is premature to pretend it can go to the core. What was your intention? How would we use it going forward? Maybe you can drop an example?

packages/platform-ui/src/pages/NotFound/NotFound.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+      a {
+        color: inherit;
+        text-decoration: none;
+      }
+    `,
+  logo: css`
+      width: 100%;
+      min-width: 250px;
+      max-width: 600px;
+    `,
```
nicolalamacchia, line 28: Spacing looks messed up here

fabio-silva: you're right 👍

tiagomotasantos (dismissed):

nicolalamacchia (approved):

atymchuk (approved):

nicolalamacchia (approved):

PR #67, opened 2021-01-13
Fix dispatch job for scheduled workflow

yurkovychv:


.github/workflows/ci.yml
```diff
@@ -128,6 +128,9 @@ jobs:
     runs-on: ubuntu-18.04
 
     steps:
+      - name: Set REPO env variable
+        run: echo "REPO=${GITHUB_REPOSITORY#*/}" >> $GITHUB_ENV
+
```
nicolalamacchia, line 133: Could you please provide more information on the problem this solves and how?

yurkovychv: Currently, we run `CI` workflow on `push`, `pull_request` and on `schedule` events. `push` and `pull_request` works great without issues. But the `schedule` event [run fails](https://github.com/percona-platform/saas-ui/runs/1676700365?check_suite_focus=true). It fails because in `Dispatch E2E tests` stage we are using `${{ github.event.repository.name }}` object, which does not exist in `schedule` event. To fix this we need to set `repository name` using this step.

atymchuk (approved):

.github/workflows/ci.yml
nicolalamacchia: Got it, thanks!

nicolalamacchia (approved):

PR #71, opened 2021-02-26
SAAS-495 auth tests

yurkovychv:


packages/platform-ui/cypress/pages/auth/utils/getNewUser.ts
```diff
@@ -1,18 +1,36 @@
 import { ValidUser } from 'pages/common/interfaces/Auth';
 import faker from 'faker';
 
+/*
+Using this lib due to a bug in a faker when passing a regex to a password function.
+https://github.com/Marak/faker.js/issues/826
+*/
+const generator = require('generate-password');
+
```
nicolalamacchia, line 9: please, import a module by using `import`

packages/platform-ui/cypress/pages/auth/utils/getNewUser.ts
```diff
@@ -1,18 +1,36 @@
...
 };
 
+export const getPassword = () => {
+  return generator.generate({
+    length: 15,
+    numbers: true,
+    lowercase: true,
+    uppercase: true,
+  });
+};
```
nicolalamacchia, line 21: do we really need a dependency to generate fake password?

yurkovychv: why shouldn't we?

packages/platform-ui/cypress/pages/auth/utils/getNewUser.ts
yurkovychv: sure

packages/platform-ui/cypress/pages/auth/utils/getNewUser.ts
atymchuk: IMHO not a problem to use an external dependency, since we:
1/ save time creating it from scratch
2/ don't want to cover it with tests
3/ don't ship it (it's not part of the bundle)

packages/platform-ui/cypress/pages/auth/utils/getNewUser.ts
```diff
@@ -1,18 +1,36 @@
...
+https://github.com/Marak/faker.js/issues/826
+*/
 
 export const getFakeEmail = () => {
   return `${faker.internet.email()}.test`;
 };
 
+export const getPassword = () => {
+  return generator.generate({
+    length: 15,
```
atymchuk, line 16: We should probably have the length as per the architecture docs, i.e. 8 characters long.

atymchuk (changes requested):

atymchuk (approved):

nicolalamacchia (approved):

PR #77, opened 2021-03-15
Experimental modification to give additional docker tag for dev-latest

browseman:


Makefile
```diff
@@ -34,7 +35,12 @@ build-ui:               ## Build platform-ui
...
+	docker build --squash --tag $(DOCKER_IMAGE):$(DOCKER_TAG) .
 
 docker-push:            ## Push Docker image
-	docker push $(DOCKER_IMAGE)
+	docker push $(DOCKER_IMAGE):$(DOCKER_TAG)
+
+docker-tag:
+	if [[ "$(DOCKER_TAG)" == "latest" ]]; then exit 0; fi; \
+	    docker tag $(DOCKER_IMAGE):$(DOCKER_TAG) $(DOCKER_IMAGE):dev-latest; \
+	    docker push $(DOCKER_IMAGE):dev-latest;
```
atymchuk, line 46:
```suggestion
	if [[ "$(DOCKER_TAG)" != "latest" ]]; then
	    docker tag $(DOCKER_IMAGE):$(DOCKER_TAG) $(DOCKER_IMAGE):dev-latest
	    docker push $(DOCKER_IMAGE):dev-latest
	fi
```
This reads better and is more straightforward )

.github/workflows/ci.yml
```diff
@@ -120,6 +120,9 @@ jobs:
       - name: Push Docker image
         run: make docker-push
 
+      - name: Tag Docker image with dev-latest
+        run: make docker-tag
```
atymchuk, line 124: Sound a bit misleading, the DOCKER_TAG is made elsewhere. We are pushing the dev-latest there.
```suggestion
        run: make docker-push-dev-latest
```

.github/workflows/ci.yml
```diff
@@ -120,6 +120,9 @@ jobs:
       - name: Push Docker image
         run: make docker-push
 
+      - name: Tag Docker image with dev-latest
```
atymchuk, line 123:
```suggestion
      - name: Tag Docker image with dev-latest and push to the registry
```

Makefile
```diff
@@ -34,7 +35,12 @@ build-ui:               ## Build platform-ui
...
 
 docker-build:           ## Build Docker image
-	docker build --squash --tag $(DOCKER_IMAGE) .
+	docker build --squash --tag $(DOCKER_IMAGE):$(DOCKER_TAG) .
 
 docker-push:            ## Push Docker image
-	docker push $(DOCKER_IMAGE)
+	docker push $(DOCKER_IMAGE):$(DOCKER_TAG)
+
+docker-tag:
```
atymchuk, line 43:
```suggestion
docker-push-dev-latest:
```

atymchuk (changes requested):

Makefile
atymchuk: It looks like we will prevent the engineers from developing in parallel, since every container in every PR will be pushed with `dev-latest` label, so the last push will overwrite the previous one. Let's discuss one more time...

nicolalamacchia: > This reads better and is more straightforward )

Added semicolons and backslashes to make this work :)

```suggestion
	if [[ "$(DOCKER_TAG)" == "latest" ]]; then \
		docker tag $(DOCKER_IMAGE):$(DOCKER_TAG) $(DOCKER_IMAGE):dev-latest; \
		docker push $(DOCKER_IMAGE):dev-latest; \
	fi; \
```

atymchuk: I guess it should be `!=` :)

browseman: Well it's better to be like this, for sure. As for why it's presented like that, because it's a leftover from previous attempt /experiment, was a bit quicker to add the exit :worried:

.github/workflows/ci.yml
browseman: Well actually it perform both tasks, 'tagging' and 'pushing' in one target inside the makefile. So not 100% sure if the naming should be `docker-tag-push-dev-latest`

Makefile
nicolalamacchia: > I guess it should be `!=`, because `== latest` was meant to exit :)

oopsie

atymchuk:
Closing this to get out of sight, we can re-open at any time.

PR #81, opened 2021-03-23
SAAS-513  UI tests for OKTA registration

yurkovychv:
tested [here](https://github.com/percona-platform/cicd/pull/28/checks?check_run_id=2174594091)

### DIFF
--- a/packages/platform-ui/cypress/integration/auth/signup.spec.ts
+++ b/packages/platform-ui/cypress/integration/auth/signup.spec.ts
@@ -59,7 +59,7 @@ context('Sign Up', () => {
       cy.mailosaurGetMessage(Cypress.env('MAILOSAUR_SAAS_SERVER_ID'), {sentTo: userEmail}, {timeout: 20000})
         .then((message) => {
           const link = message.html!.links!
-            .find(({ text }) => text === '  Activate Okta Account  ')!.href;
+            .find(({ text }) => text!.trim() === 'Activate Okta Account')!.href;
 
           cy.visit(link!);
           cy.get('[name="newPassword"]').type(userPassword);
--- a/packages/platform-ui/cypress/integration/auth/signup.spec.ts
+++ b/packages/platform-ui/cypress/integration/auth/signup.spec.ts
@@ -1,47 +1,88 @@
 /// <reference types="cypress" />
-import { INVALID_USER, pageDetailsMap, Pages, VALID_USER } from 'pages/common/constants';
+import { EXISTING_USER, INVALID_USER, pageDetailsMap, Pages } from 'pages/common/constants';
 import { runFieldsValidationFlow } from 'pages/auth/flows/validation.flow';
 import { runPageElementsFlow } from 'pages/auth/flows/checkElements.flow';
-import { runSignUpFlow } from 'pages/auth/flows/auth.flow';
-import { getNewUser } from 'pages/auth/utils/getNewUser';
-import { runLoginAction } from 'pages/auth/actions/login.action';
-import { profileIcon, userEmail } from 'pages/main/view/selectors';
+import { runLoginFlow, runSignUpFlow } from 'pages/auth/flows/auth.flow';
+import { getUser } from 'pages/auth/utils/getUser';
 import { setAliases } from 'pages/auth/requests/requests';
 import { popUp } from 'pages/common/view/selectors';
-import { loginForm, signupForm, signUpLink } from 'pages/auth/view/selectors';
+import { loginForm, signupForm, signUpLink, submitButton } from 'pages/auth/view/selectors';
 import { runSignupAction } from 'pages/auth/actions/signup.action';
+import { fillEmailPassword } from 'pages/auth/view/behavior/auth';
 
-const newUser = getNewUser();
+const newUser = getUser();
 
 context('Sign Up', () => {
   beforeEach(() => {
     setAliases();
-    cy.visit(pageDetailsMap[Pages.SignUp].url);
   });
 
   it('SAAS-T82 - should be able to see the signup form', () => {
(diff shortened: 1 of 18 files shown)


packages/platform-ui/cypress/integration/auth/signup.spec.ts
```diff
@@ -1,47 +1,88 @@
...
+import { getUser } from 'pages/auth/utils/getUser';
 import { setAliases } from 'pages/auth/requests/requests';
 import { popUp } from 'pages/common/view/selectors';
-import { loginForm, signupForm, signUpLink } from 'pages/auth/view/selectors';
+import { loginForm, signupForm, signUpLink, submitButton } from 'pages/auth/view/selectors';
 import { runSignupAction } from 'pages/auth/actions/signup.action';
+import { fillEmailPassword } from 'pages/auth/view/behavior/auth';
 
-const newUser = getNewUser();
+const newUser = getUser();
```
atymchuk, line 13: One potential problem area I see is that `newUser` is being modified between different unit tests, which is a bit against one of the requirements, which says that different unit tests should not modify common state, unless it's absolutely unavoidable. Not really sure about the execution sequence in `cypress` (I guess it is not parallel), but doing this way may not guarantee the realiability.

This is just a general observation to consider for the future iterations, not necessarily a requirement to change the tests here if they pass.

packages/platform-ui/cypress/support/commands.js
```diff
@@ -1,4 +1,5 @@
 import '@testing-library/cypress/add-commands';
+import 'cypress-mailosaur';
```
atymchuk, line 2: I understand we are not using mailosaur in this PR, right? If so, maybe consider adding it later?

atymchuk (approved):

packages/platform-ui/cypress/support/commands.js
yurkovychv: We actually use it [here](https://github.com/percona-platform/saas-ui/blob/327c8a212d5d20e59f3c9be9f93652dcaff4e091/packages/platform-ui/cypress/integration/auth/signup.spec.ts#L59) to get activation link.

packages/platform-ui/cypress/integration/auth/signup.spec.ts
yurkovychv: I see your point, will try to search for a solution

packages/platform-ui/cypress/integration/auth/signup.spec.ts
```diff
@@ -1,47 +1,88 @@
...
+      loginForm().isVisible();
+    });
+  });
+
+  it('should be able to activate account', { baseUrl: null}, () => {
+    cy.task('getUser').then(({ userEmail, userPassword }) => {
+      cy.mailosaurGetMessage(Cypress.env('MAILOSAUR_SAAS_SERVER_ID'), {sentTo: userEmail}, {timeout: 20000})
+        .then((message) => {
+          const link = message.html!.links!
+            .find(({ text }) => text === '  Activate Okta Account  ')!.href;
```
nicolalamacchia, line 62: I think this will change in the (near?) future

nicolalamacchia (approved):

packages/platform-ui/cypress/integration/auth/signup.spec.ts
yurkovychv: will update it once it's changed

atymchuk: Much more reliable to trim the variable and compare it to the constant `Activate Okta Account`, cause if I change the template it will break :)

packages/platform-ui/cypress/integration/auth/signup.spec.ts
atymchuk: > I see your point, will try to search for a solution

Well, `beforeEach` can be our friend, i.e. you can create the user object for every unit.

yurkovychv: The thing is that I don't want to have a user object for every unit. I was fighting for having the user shared across some tests. One of Cypress's weaknesses is cross-origin. So had to find some workaround

packages/platform-ui/cypress/integration/auth/signup.spec.ts
yurkovychv: you are right. Thanks. Will do

packages/platform-ui/cypress/integration/auth/signup.spec.ts
atymchuk: I was talking about the units, not suites. Are we on the same page?

packages/platform-ui/cypress/support/commands.js
atymchuk: Oh, missed it. Makes sense now )

packages/platform-ui/cypress/integration/auth/signup.spec.ts
yurkovychv: yeah, we are. I'm sharing a generated email and password between few test units.

atymchuk: Sharing is OK, modifying is not.

packages/platform-ui/cypress/integration/auth/signup.spec.ts
yurkovychv: Done

packages/platform-ui/cypress/integration/auth/signup.spec.ts
atymchuk: Closing the thread, since it seems the only way we can do it right now.

PR #91, opened 2021-05-14
PMM-6627 trim input fields

nicolalamacchia:


tiagomotasantos (approved):

fabio-silva (approved):
Seems OK. Just one question: since we are using the `required` attribute along with validators, does that mean the user will see both the native error tooltip and our own error message below the field?

nicolalamacchia:
@fabio-silva that required is not used on native elements, so no :)

PR #98, opened 2021-06-29
Add CodeQL in CI

nicolalamacchia:


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,29 @@
+name: "CodeQL"
+
+on: [push]
```
atymchuk, line 3: I assume this workflow is quite expensive. Do we want to run it on every push, or would it suffice to run on push to main only?

nicolalamacchia: It takes less than 3 mins to run, in parallel with the other workflows, I think we can keep it to have more feedbacks in PRs as well

tiagomotasantos (dismissed):

fabio-silva (dismissed):

.github/workflows/codeql.yml
nicolalamacchia: updated to run only on main and pull requests

tiagomotasantos (approved):

fabio-silva (approved):

atymchuk (approved):

PR #99, opened 2021-07-08
SAAS-548 update profile tests

yurkovychv:
[SAAS-561 PR](https://github.com/percona-platform/saas-ui/pull/96) is included here.


package.json
```diff
@@ -73,6 +73,7 @@
     "@testing-library/cypress": "^7.0.4",
     "@types/faker": "^5.1.2",
     "@types/react-redux": "^7.1.9",
+    "axios": "^0.21.1",
```
nicolalamacchia, line 76: @yurkovychv why do we need `axios`? Isn't it possible to use `fetch` as we already do?

yurkovychv: @nicolalamacchia I think I don't get your point. There is no `fetch` in dependencies. Could you point me to where `fetch` is being used?

puneet0191 (approved):

package.json
nicolalamacchia: `fetch` is native, no need for a library :)

yurkovychv: does not work with node application) that's why I asked )

nicolalamacchia: oh man, you're right, sorry for the confusion 🤦 I'm too used to front end code 😅

nicolalamacchia (approved):

cypress/plugins/oktaRequest.js
```diff
@@ -0,0 +1,37 @@
+const axios = require('axios');
+
+export const oktaRequest = async ({baseUrl, urlSuffix, method = 'get', token, data = {}}) => {
+  let response;
+
+  try {
```
atymchuk, line 6: If we leverage a promise syntax here, i.e. `.then()` and `.catch`, it could look much cleaner.

atymchuk (approved):

cypress/plugins/oktaRequest.js
yurkovychv: thanks, will do next time

PR #107, opened 2021-08-30
SAAS-632 Update org creation status if orgs exist

nicolalamacchia:


src/pages/GettingStarted/GettingStarted.tsx
```diff
@@ -1,20 +1,50 @@
...
+  useEffect(() => {
+    if (error) {
+      toast.error(Messages.fetchError);
+    }
+  }, [error]);
+
+  useEffect(() => {
+    const { org_ids } = data;
+
+    setOrgIds(org_ids);
```
atymchuk, line 26:
```suggestion
    setOrgIds(data?.org_ids);
```
No need to destructure IMO )

atymchuk (approved):

src/pages/GettingStarted/GettingStarted.tsx
nicolalamacchia: Good point, done

fabio-silva (approved):

PR #109, opened 2021-09-10
SAAS-633 show org details

nicolalamacchia:
image
image


src/pages/GettingStarted/GettingStarted.tsx
```diff
@@ -12,7 +12,7 @@ import { LIST_ORGANIZATION_URL } from './GettingStarted.constants';
 export const GettingStartedPage: FC = () => {
   const styles = useStyles(getStyles);
   const [orgIds, setOrgIds] = useState([]);
-  const { error, data = {} } = useFetch(LIST_ORGANIZATION_URL, { method: 'POST' }, []);
+  const { error, data = {} } = useFetch(`http://localhost:5000/${LIST_ORGANIZATION_URL}`, { method: 'POST' }, []);
```
fabio-silva, line 15: Why are we hardcoding localhost and port here?

src/pages/ManageOrganization/OrganizationTab/OrganizationView/OrganizationView.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  useEffect(() => {
+    if (error) {
+      toast.error(Messages.fetchError);
+    }
+  }, [error]);
+
+  useEffect(() => {
+    setOrgName(data?.name);
+    setOrgCreationDate(new Date(data?.created_at).toLocaleDateString());
```
fabio-silva, line 25: Do we want to set the date without doing any checks? This can easily go "Invalid Date", which means the check you have on the render (`orgName && orgCreationDate`) can cause a false positive, even though the date isn't valid.

src/pages/ManageOrganization/OrganizationTab/OrganizationTab.tsx
```diff
@@ -1,54 +1,30 @@
...
-              loading={loading}
-              disabled={!valid || loading || pristine}
-            >
-              {Messages.save}
-            </LoaderButton>
-          </form>
-        )}
-      </Form>
+    <div data-qa="manage-organization-organization-tab">
+      {orgId!! ? <OrganizationView orgId={orgId!} /> : <OrganizationCreate />}
```
fabio-silva, line 27: Isn't one exclamation mark enough?
Anyway, if we are adding `!!` to tell the compiler the var can't be null/undefined, why the ternary?
Shouldn't it be `{orgId ? <OrganizationView orgId={orgId} /> : <OrganizationCreate />}` ?

fabio-silva (changes requested):

src/pages/GettingStarted/GettingStarted.tsx
nicolalamacchia: Oops, fixed it!

src/pages/ManageOrganization/OrganizationTab/OrganizationTab.tsx
nicolalamacchia: No, it's not enough because orgId could be undefined, but by checking that it exists before (ternary condition) we can safely tell TypeScript that `orgId` in the params is not undefined. The `!!` is not telling the compiler it can't be null/undefined, it's casting to boolean (which as a side effect not to be null/undefined), to be explicit the second `!` in params is needed.

src/pages/ManageOrganization/OrganizationTab/OrganizationView/OrganizationView.tsx
nicolalamacchia: We trust the back end to send correct data, the creation date must exist by requirement, so it's ok to just convert it

src/pages/ManageOrganization/OrganizationTab/OrganizationTab.tsx
fabio-silva: I see, but then I guess it should be `!!orgId`, right? And by doing that, the exclamation mark on the left-side expression can be removed.

src/pages/ManageOrganization/OrganizationTab/OrganizationView/OrganizationView.tsx
fabio-silva: Roger 👍

src/pages/ManageOrganization/OrganizationTab/OrganizationCreate/OrganizationCreate.messages.ts
```diff
@@ -0,0 +1,8 @@
+export const Messages = {
+  createOrganization: 'Create Organization',
+  orgNamePlaceholder: 'Your organization name',
```
tiagomotasantos, line 3: Some word start with capital letters while others don't. Can we change too "Create organization" and "Organization name" or does it really need to be like this?

src/pages/ManageOrganization/OrganizationTab/OrganizationCreate/OrganizationCreate.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }, [error, data]);
+
+  return (
+    <div data-qa="create-organization-wrapper" className={styles.container}>
+      <OrganizationLogo />
+      <h4 className={styles.title}>{Messages.createOrganization}</h4>
+      <Form onSubmit={handleCreateOrgSubmit}>
+        {({ handleSubmit, pristine, valid }: FormRenderProps) => (
+          <form data-qa="create-organization-form" className={styles.form} onSubmit={handleSubmit}>
+            <TextInputField className={styles.orgNameInput} label={Messages.organizationName} placeholder={Messages.orgNamePlaceholder} name="organizationName" validators={[validators.required]} />
```
atymchuk, line 40: This needs to be wrapped, like the Loader is below

src/pages/ManageOrganization/OrganizationTab/OrganizationTab.tsx
atymchuk: Agree, looks strange. Can you provide any reference to TS docs to see if this is known/best practice/etc?

nicolalamacchia: oh, sorry, now I get it, you mean that `!!` is redundant, right? yes, it's just more explicit, but I consider this a nit. The second one is still needed, otherwise TS complains.

src/pages/ManageOrganization/OrganizationTab/OrganizationCreate/OrganizationCreate.messages.ts
nicolalamacchia: @tiagomotasantos because some of them are titles and some other are not

src/pages/ManageOrganization/OrganizationTab/OrganizationCreate/OrganizationCreate.tsx
nicolalamacchia: Can you please explain more in details what you mean?

nicolalamacchia: done

src/pages/ManageOrganization/OrganizationTab/OrganizationTab.tsx
nicolalamacchia: done

tiagomotasantos (approved):

PR #114, opened 2021-10-05
SAAS-642 invite members

nicolalamacchia:


src/pages/ManageOrganization/ManageOrganization.tsx
```diff
@@ -1,14 +1,80 @@
...
+    if (org?.id && response.ok) {
+      toast.success(Messages.orgCreateSuccess);
+      setOrgId(org?.id);
+    }
+  }, [post, response.ok]);
+
+  const tabs = useMemo(() => [
+    {
+      label: Messages.members,
+      disabled: !orgId,
```
fabio-silva, line 33: any chance the first ever org ID might be a 0 and this becomes disabled?

fabio-silva (dismissed):

src/pages/ManageOrganization/ManageOrganization.tsx
nicolalamacchia: good point, the type of `orgId` was actually wrong, since it's a UUID (a string). fixed the types

fabio-silva (approved):

src/pages/ManageOrganization/InviteMember/InviteMember.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { TestContainer } from 'components/TestContainer';
+import { InviteMember } from '.';
+
+const mockPost = jest.fn().mockResolvedValue({ orgs: [{ id: 1337 }] });
+
+const toastError = jest.spyOn(toast, 'error');
+
+let mockError: string | null = null;
+
+jest.mock('use-http', () => {
```
tiagomotasantos, line 13: nit: maybe we can find a way to put this in a function or something to avoid writing it in every test that uses `useFetch`

src/pages/ManageOrganization/InviteMember/InviteMember.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const submitButton = await screen.findByTestId('invite-member-submit-button');
+    const emailInput = await screen.findByTestId('email-text-input');
+
+    fireEvent.change(emailInput, { target: { value: 'invalid@email.com' } });
+
+    expect(submitButton).toBeEnabled();
+  });
+
+  test('shows an error if an API call fails', async () => {
+    mockError = 'Error';
```
tiagomotasantos, line 132: I don't understand how this mockError gets used. Anyway in this case maybe it doesn't even need to be a variable

src/pages/ManageOrganization/InviteMember/InviteMember.tsx
```diff
@@ -0,0 +1,92 @@
...
+      </Button>
+      <Modal
+        title={Messages.inviteMember}
+        isVisible={isModalVisible}
+        onClose={handleModalClose}
+      >
+        <Form onSubmit={handleInviteMemberSubmit} initialValues={initialValues}>
+          {({ handleSubmit, valid, pristine }: FormRenderProps) => (
+            <form onSubmit={handleSubmit} className={styles.inviteForm} data-testid="invite-member-form">
+              <TextInputField name="email" label={Messages.email} validators={[emailValidator, required]} data-testid/>
```
tiagomotasantos, line 68: Empty testid

src/pages/ManageOrganization/InviteMember/InviteMember.test.tsx
nicolalamacchia: it changes the value of the `mockError` variable defined on the topmost scope.

.eslintrc
```diff
@@ -82,7 +82,8 @@
     ],
     "no-inline-styles/no-inline-styles": ["warn"],
     "max-classes-per-file": ["error"],
-    "no-shadow": ["error"],
+    "no-shadow": ["off"],
```
atymchuk, line 85: Why? Relaxing a rule like this may lead to all sorts of surprises. Is there a strong reason?

src/pages/ManageOrganization/InviteMember/InviteMember.constants.ts
```diff
@@ -0,0 +1,17 @@
+import { Messages } from './InviteMember.messages';
+import { MemberRole } from './InviteMember.types';
+
+export const GET_ORGANIZATION_URL = 'v1/orgs';
```
atymchuk, line 4: We already have a place for all urls in endpoints.ts

src/pages/ManageOrganization/InviteMember/InviteMember.messages.ts
```diff
@@ -0,0 +1,10 @@
+export const Messages = {
+  admin: 'Admin',
+  email: 'Email',
+  fetchError: 'An error occurred, please try again.',
```
atymchuk, line 4: This message should be more specific, i.e. something like 'Couldn't fetch the member(s)'

src/pages/ManageOrganization/InviteMember/InviteMember.messages.ts
```diff
@@ -0,0 +1,10 @@
...
+  inviteMember: 'Invite Member',
+  inviteSuccess: 'User invited successfully',
```
atymchuk, line 6:
```suggestion
  inviteSuccess: 'User successfully invited',
```

.eslintrc
nicolalamacchia: it's a known bug, it doesn't work with enums in TypeScript. You can google "typescript enum no-shadow" for more info. It's not off anyway, it has been replaced with `"@typescript-eslint/no-shadow": ["error"],`, which works the same way but doesn't trigger a false-positive

src/pages/ManageOrganization/InviteMember/InviteMember.test.tsx
nicolalamacchia: will try to find a way in the next tickets

tiagomotasantos (approved):

src/pages/ManageOrganization/InviteMember/InviteMember.messages.ts
atymchuk: "The use has been successfully invited." Is it more clear now?

atymchuk: The order of the words should be respected when you omit other elements.

tiagomotasantos (dismissed):

PR #121, opened 2021-11-05
Okta: get user info into store

fabio-silva:
- Cleaned login / logout unused actions
- `useUserInfo` hook takes care of calling Okta's endpoint and fill in the store the first time


src/core/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,42 @@
+import { useCallback, useEffect } from 'react';
+import { useOktaAuth } from '@okta/okta-react';
+import { useDispatch, useSelector } from 'react-redux';
+import { AuthState, UpdateProfilePayload } from 'store/types';
+import { RequestError } from 'core/api/types';
+import { authGetProfileAction, authUpdateProfileAction, getAuth } from '../../store/auth';
```
atymchuk, line 6:
```suggestion
import { authGetProfileAction, authUpdateProfileAction, getAuth } from 'store/auth';
```
nit

src/core/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const useUserInfo = (): [AuthState, (payload: UpdateProfilePayload) => void] => {
+  const { oktaAuth } = useOktaAuth();
+  const dispatch = useDispatch();
+  const user = useSelector(getAuth);
+  const { freshData } = user;
```
atymchuk, line 12: `freshData` reads somewhat weird )
Is it us or Okta who came up with that prop name?

src/store/auth/auth.reducer.ts
```diff
@@ -1,9 +1,10 @@
 import { createAsyncAction, ActionType, getType } from 'typesafe-actions';
-import { AuthState, LoginPayload, LogoutPayload, UpdateProfilePayload } from 'store/types';
+import { AuthState, UpdateProfilePayload } from 'store/types';
 import { RequestError } from 'core/api/types';
 
 const DEFAULT_STATE: AuthState = {
   authenticated: false,
+  freshData: true,
```
atymchuk, line 7: How can `freshData` be true by default? Not sure it's even needed )

src/core/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+  }, [dispatch]);
+
+  useEffect(() => {
+    const getInfo = async() => {
+
+      if (freshData) {
+        dispatch(authGetProfileAction.request());
+
+        try {
+          const { email: oktaEmail, family_name, given_name } = await oktaAuth.getUser();
```
atymchuk, line 25:
```suggestion
          const { email, family_name, given_name } = await oktaAuth.getUser();
```

src/core/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+ 
+          dispatch(authGetProfileAction.success({
+            email: oktaEmail,
```
atymchuk, line 28:
```suggestion
            email,
```

src/core/hooks/useUserInfo.ts
atymchuk: No need to be renamed, the email is not Okta's.

atymchuk (changes requested):

nicolalamacchia (dismissed):
just found some nits @atymchuk already pointed to

src/core/hooks/useUserInfo.ts
fabio-silva: 👍

src/core/hooks/useUserInfo.ts
fabio-silva: removed flag

src/core/hooks/useUserInfo.ts
fabio-silva: I renamed this var. It must have another name, as we already got `email` some lines above (`const { email } = user;`)

src/core/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+  }, [dispatch]);
+
+  useEffect(() => {
+    const getInfo = async() => {
+
+      if (!email) {
+        dispatch(authGetProfileAction.request());
+
+        try {
+          const { email: oktaUserEmail, family_name, given_name } = await oktaAuth.getUser();
```
atymchuk, line 25:
```suggestion
          const { email, family_name, given_name } = await oktaAuth.getUser();
```

src/core/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+ 
+          dispatch(authGetProfileAction.success({
+            email: oktaUserEmail,
```
atymchuk, line 28:
```suggestion
            email,
```

src/core/hooks/useUserInfo.ts
atymchuk: it's a nit, but why do we have to attach the word "okta" to the properties? It is a user's email after all

atymchuk (changes requested):

src/core/hooks/useUserInfo.ts
fabio-silva: good one 👍

src/core/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+        } catch (e) {
+          dispatch(authGetProfileAction.failure(e as RequestError));
+        }
+      }
+    };
+
+    getInfo();
+  }, [oktaAuth, dispatch, email]);
+
+  return [{ ...user }, setUser];
```
atymchuk, line 41: No need to copy imho, it will always return a new reference whenever things change in the user's state

atymchuk (approved):

PR #130, opened 2021-11-17
SAAS-744 logout error message

nicolalamacchia:


fabio-silva (dismissed):

tiagomotasantos (dismissed):

atymchuk:
@nicolalamacchia This is a bug, and there is a ticket open. Let's stick to the branch naming convention.

src/components/MenuBar/MenuBar.tsx
```diff
@@ -31,7 +31,8 @@ export const MenuBar: FC = () => {
     dispatch(authLogoutAction.request());
     try {
       await oktaAuth.signOut({ revokeAccessToken: true, revokeRefreshToken: true });
-      dispatch(authLogoutAction.success());
```
atymchuk, line 34: This is not the right cure, because if we do not dispatch the `authLogoutAction.success()` action, the redux state is going to have `authenticated: true`, which is absolutely not the case (see [here](https://github.com/percona-platform/saas-ui/blob/main/src/store/auth/auth.reducer.ts#L72-L78)).

Solution: we should rely on the `authenticated` state from Okta auth, rather than our own.

atymchuk (changes requested):

src/components/MenuBar/MenuBar.tsx
atymchuk: I suggest that [this](https://github.com/percona-platform/saas-ui/blob/5a7231094f603905a0074d6c0371caa267ed0221/src/store/types/auth.ts#L3) be removed from the store, it's a relatively small refactoring.

nicolalamacchia: @atymchuk if you inspect the state with redux devtools, you will see that the state gets reset after logout, because of the refresh. so the `authenticated` state will be `false`.

nicolalamacchia: That's not the right it's a hotfix which will need a proper fix

atymchuk: > @atymchuk if you inspect the state with redux devtools, you will see that the state gets reset after logout, because of the refresh. so the `authenticated` state will be `false`.

Do we want to rely on a side effect? How can one reason reason about it going forward?

nicolalamacchia: I agree it's confusing, but this is supposed to be a quick fix for the demo. Again, it will require a proper fix and investigation which needs to be prioritized. If there will be some refactoring needed as a result of that investigation, it will be addressed.

PR #134, opened 2021-11-22
SAAS-744 fix logout

nicolalamacchia:


fabio-silva (approved):

solovevayaroslavna (approved):

atymchuk (approved):
Great cleanup! Do we need to keep the email in the store once the user logs out? It looks like we should remove it as well, no?

nicolalamacchia:
> Great cleanup! Do we need to keep the email in the store once the user logs out? It looks like we should remove it as well, no?

no, there's a full refresh and the store is cleaned up

PR #147, opened 2021-11-30
Update GettingStartedOrgSection.messages.ts

rnovikovP:


rnovikovP:
@nicolalamacchia  PTAL, is this the correct place to suggest changes?

nicolalamacchia:
> @nicolalamacchia PTAL, is this the correct place to suggest changes?

image

It looks a bit too long to me. Do we want such a long text in that place?

atymchuk:
@rnovikovP Seems quite stale and outdated... how about closing it?

PR #151, opened 2021-12-03
SAAS-740 Adds dynamic contacts for Percona customers

tiagomotasantos:
Adds dynamic contacts from Customer Success Manager for Percona customers

Screenshot 2021-12-03 at 15 30 10

https://jira.percona.com/browse/SAAS-740


src/pages/Dashboard/Contacts/CustomerContact/CustomerContact.tsx
```diff
@@ -0,0 +1,47 @@
...
+  name,
+  email,
+  ticketUrl,
+}) => {
+  const styles = useStyles(getStyles);
+  const createTicket = useCallback(() => {
+    window.open(ticketUrl, '_blank', 'noopener,noreferrer');
+  }, [ticketUrl]);
+  const copyToClipboard = useCallback(() => {
+    navigator.clipboard.writeText(email);
```
atymchuk, line 18: Don't we have a standard component to copy text in core?

src/pages/Dashboard/Contacts/CustomerContact/CustomerContact.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { Messages } from './CustomerContact.messages';
+
+export const CustomerContact: FC<CustomerContactProps> = ({
+  name,
+  email,
+  ticketUrl,
+}) => {
+  const styles = useStyles(getStyles);
+  const createTicket = useCallback(() => {
+    window.open(ticketUrl, '_blank', 'noopener,noreferrer');
```
atymchuk, line 15: Can this be a function in tools to avoid writing boilerplate like `noopener,noreferrer`?

src/pages/Dashboard/Contacts/CustomerContact/CustomerContact.tsx
tiagomotasantos: Yes but that component is a button and we want an icon. I tried to use it but the styles are very different and it's much easier to just add this line then restyle the component.

src/pages/Dashboard/Contacts/CustomerContact/CustomerContact.tsx
tiagomotasantos: Done!

src/pages/Dashboard/Contacts/CustomerContact/CustomerContact.tsx
atymchuk: Actually, that was meant to be an icon, since we already have a button in grafana/ui.
@solovevayaroslavna can you help figure out if that brand new component can be used here?

nicolalamacchia (dismissed):

src/pages/Dashboard/Contacts/Contacts.tsx
```diff
@@ -1,21 +1,92 @@
...
+  const [sucessManager, setSuccessManager] = useState<SuccessManager>();
+  const [isCustomer, setIsCustomer] = useState(false);
   const [user] = useUserInfo();
   const [role, rolePending] = useUserRole();
+  const fetchConfig = getUseHttpConfig();
+  const { error, loading, post, get } = useFetch(...fetchConfig);
+  const {
+    loading: loadingCompany,
+    post: postUserCompany,
+  } = useFetch(...getUseHttpConfig(undefined, { loading }));
```
nicolalamacchia, line 31: Why do you use another `useFetch`? You should use the `post`/`get` functions from the first one.

nicolalamacchia (approved):

src/pages/Dashboard/Contacts/Contacts.tsx
tiagomotasantos: The reason is that we don't want to handle the error done with the second useFetch, if ir fails we don't want to show a message. While the first request is the opposite. Also we use the loaders from both requests in different places and we dont want one to affect the other

solovevayaroslavna (approved):

PR #160, opened 2021-12-15
SAAS-794 migrate to redux toolkit

atymchuk:
Notable things:
* typescript downgraded to match the core-ui
* devtools are now logging actions to the console (it's configurable)
* migrated most of the component-based API calls to thunks
* moved the initial okta.getUser call to a middleware to guarantee the presence of authenticated user in the store
* store orgs, members and pretty much everything we get from API in redux
* migrated API calls chaining logic to the thunks
* removed demo ui, i.e. the `/ui` route and components

TODO: fix unit tests

P.S. let's use the logger as much as we can, since it is not polluting the screen when running in unit tests


atymchuk:
@nicolalamacchia Awesome, LGTM.
Let's se what we can do with unit tests.

fabio-silva (changes requested):
Just some things:
- Some errors along the code, regarding unused/non-defined vars;
- We are excluding some tests. Is there a reason for that?

src/pages/Login/Login.test.tsx
```diff
@@ -15,7 +15,7 @@ jest.mock('@okta/okta-react', () => ({
   }),
 );
 
-describe('Portal Login', () => {
+xdescribe('Portal Login', () => {
```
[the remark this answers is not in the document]
nicolalamacchia: > Just some things:
> 
> * Some errors along the code, regarding unused/non-defined vars;
> * We are excluding some tests. Is there a reason for that?

Yes, we are excluding that because they all need to be upgraded, but it's better to do so after the implementation of a possible new way of making API calls, which will come separately.

src/store/auth/auth.actions.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    // We do not show the toast on logout success
+    return;
+  } catch (err) {
+    displayAndLogError(err);
+  }
+});
+
+// USER:INFO actions are intentinally synchronous, since they are part of the middleware
+export const startUserInfoAction = createAction<void>('USER:INFO/START');
```
tiagomotasantos, line 43: Why are user related actions on the auth file? Maybe we should leave only login/logout here and move user actions to another file (e.g. user.actions)?

src/store/auth/auth.actions.ts
```diff
@@ -0,0 +1,103 @@
...
+
+      return rejectWithValue(err);
+    }
+  },
+);
+
+export const updateProfileAction = createAsyncThunk<
+  UpdateProfilePayload,
+  UpdateProfilePayload,
+  { rejectValue: unknown }
```
tiagomotasantos, line 70: Does the API return the errors in a specific format? If so we could create a type for errors here

nicolalamacchia: @atymchuk did you look into that already?

src/store/auth/auth.actions.ts
nicolalamacchia: I guess login/sign-up/logout are still use related, IMO you have a good point about the name though. Although this is a nit, we can consider changing the name. cc @atymchuk

tiagomotasantos (dismissed):

fabio-silva (approved):

tiagomotasantos (approved):

fabio-silva (dismissed):

fabio-silva (approved):

fabio-silva (approved):

PR #161, opened 2021-12-16
SAAS-791 Improve entitlements view

tiagomotasantos:
Improves entitlements view:

- Makes the modal header bigger
- Fixes the modal size so it doesn't grow/shrink
- Adds expiry date

Expiry date was added through css because in the version of Grafana we are using the Collapse component only allows a single text but in Grafana 8 the component already supports a ReactNode as header so we can refactor this part when we update to Grafana 8 to cleanup the css.

https://jira.percona.com/browse/SAAS-791

https://user-images.githubusercontent.com/16366502/146408221-c885569a-852b-4ac3-bfe4-15d71575d826.mov


src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Messages } from './EntitlementsModal.messages';
+
+const NEXT_CHILD = 1;
+
+// TODO: we can remove the styles for the expiry date in the collapse after
+// updating to Grafana 8 because the Collapse component already allows a ReactNode
+// to be used as header there
+export const getStyles = ({ palette, typography, spacing }: GrafanaTheme) => ({
+  modalWrapper: (entitlements: OrganizationEntitlement[]) => css`
+    div[data-testid="modal-body"] {
```
atymchuk, line 13: @tiagomotasantos Let's see if can have a different selector so we do not depend on the test-id, which may change and we won't probably notice.

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+    div[data-testid="modal-content"] {
+      max-height: calc(90vh - ${spacing.xl});
+    }
+
+    div[class$="collapse__header-label"] {
+      position: relative;
+    }
+
+    ${entitlements.map(({ end_date }, i) => (
+      `div[data-testid="modal-content"] {
```
atymchuk, line 35: Is there no way to reach to those elements other than checking all the nested divs? This code doesn't look much like maintainable :)

atymchuk (changes requested):

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
tiagomotasantos: There's no different selector unfortunately. Our modal from platform core doesn't allow to receive a className to do this. And even then we would have to play with the structure of the elements inside the modal to change width/height. The best we can do here is improve the API of the modal to handle this cases.

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
tiagomotasantos: There's no other way at the moment. I left a comment in the PR and in the code regarding this.
We need to add the expiry date to the collapse header, in the version of grafana we are using it only allows a string as header. In version 8 already allows a ReactNode, so when we upgrade to version 8 we can remove this code and use a custom header.

(and I also tried to do this through the content of the collapse but it gets removed from the DOM when the collapse if closed and we need this to be always shown so that's not even a option)

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
tiagomotasantos: (of course we could use the order of the elements to get to the correct element but I think that's much worse than using testid)

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.tsx
```diff
@@ -16,43 +16,47 @@ export const EntitlementsModal: FC<EntitlementsModalProps> = ({ entitlements, on
...
+                <p><span>{Messages.startDate}:</span>{new Date(start_date).toLocaleDateString()}</p>
+                <p><span>{Messages.endDate}:</span>{new Date(end_date).toLocaleDateString()}</p>
+                <p><span>{Messages.summary}:</span>{summary}</p>
+                <p><span>{Messages.tier}:</span>{tier}</p>
+                <p><span>{Messages.totalUnits}:</span>{total_units}</p>
+                <p><span>{Messages.supportLevel}:</span>{support_level}</p>
+                <p><span>{Messages.platform}</span></p>
+                <div className={styles.advisorsWrapper} data-testid="entitlements-advisors-wrapper">
+                  <Advisor label={Messages.configAdvisor} hasAvisor={config_advisor} />
+                  <Advisor label={Messages.securityAdvisor} hasAvisor={security_advisor} />
```
fabio-silva, line 53: nit: is `hasAvisor` a typo?

fabio-silva (dismissed):

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.tsx
tiagomotasantos: Totally! Thank you, fixed!

atymchuk (approved):

fabio-silva (approved):

PR #163, opened 2021-12-23
SAAS-709 UI tests for new flow

yurkovychv:


package.json
```diff
@@ -78,7 +79,7 @@
     "@testing-library/cypress": "^7.0.4",
     "@types/faker": "^5.1.2",
     "@types/react-redux": "^7.1.9",
-    "cypress": "^7.3.0",
+    "cypress": "^9.2.0",
     "cypress-cy-select": "^0.2.2",
```
atymchuk, line 83: What about upgrading the rest of cypress-related packages?

atymchuk (approved):

puneet0191 (approved):

package.json
yurkovychv: Sorry, missed it. Should I do it in the scope of this PR?

tiagomotasantos (approved):

package.json
atymchuk: Not necessarily, up to you. Just trying to make sure we keep all of the cypress deps updated.

PR #165, opened 2021-12-29
SAAS-688 Add Google Analytics code

nicolalamacchia:


public/index.html
```diff
@@ -12,6 +12,13 @@
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <title>Percona Platform Portal</title>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-343802-23"></script>
```
atymchuk, line 15: Can we disable it in the dev environment?

nicolalamacchia: @rnovikovP told me he wanted to categorize events based on the domain (dev vs prod), so I guess it's better to keep it. Let's wait for him to confirm.

atymchuk: That is surprising, never heard of the need to analyze the dev.

fabio-silva (approved):

public/index.html
```diff
@@ -12,6 +12,28 @@
...
+      }
+
+      head.appendChild(js);
+    </script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      if (window.location.hostname === '%REACT_APP_PROD_DOMAIN%') {
+        gtag('config', 'UA-343802-23');
```
tiagomotasantos, line 32: Can this config be an env variable?
For example .env file contains prod tag UA-343802-23 while .env.development contains dev tag UA-343802-16.
In that case there's no need for conditions neither to have REACT_APP_PROD_DOMAIN

tiagomotasantos (changes requested):

public/index.html
nicolalamacchia: @tiagomotasantos those variables are replaced at build time, which is not something we want right now (it would mean two builds, one per environment).

tiagomotasantos: `npm run build` creates a prod build already. We would have to find a way just to create a dev build for local development. Why we don't want that? 🤔

In any case if that's not an option I don't see the need for REACT_APP_PROD_DOMAIN, we could ate least use NODE_ENV (set it to prod where we are building it for prod and the rest of the environments can even be unset)

nicolalamacchia: We would have to maintain two builds, one for the dev env and one for prod, which is something we discussed already with the team and found it's an overkill, since we'd need it just in a couple of places, where we can just use the check against the domain name at runtime (which is a common strategy).

The NODE_ENV variable gets replaced at build time as well, of course, and it requires a specific build command for dev, which we don't want for what I've mentioned before.

Consider that the dev build is not needed for local development, but for the dev environment, so it has to be pushed to the registry and requires CI to be aware of it.

tiagomotasantos: Got it, thanks!

tiagomotasantos (approved):

PR #169, opened 2022-01-13
SAAS-816 create organizations tests

yurkovychv:


cypress/integration/customer/organizations/helper.js
```diff
@@ -0,0 +1,31 @@
...
+  cy.findAllByTestId('table-tbody-tr').should('be.visible');
+};
+
+export const prepareOrganizationWithAdminAndTechnical = (snAccount) => {
+  // create users
+  cy.oktaCreateUser(snAccount.admin1);
+  cy.oktaCreateUser(snAccount.technical);
+
+  // create org by admin
+  cy.getUserAccessToken(snAccount.admin1.email, snAccount.admin1.password).then((adminAccessToken) => {
```
nicolalamacchia, line 23: nit: to avoid nesting and to make this more readable, you could make this function `async`, use `await` here and save the result in a const (`adminAccessToken`) and then `cy.apiCreateOrg(adminAccessToken);`.

The same applies to other places in this PR.

cypress/integration/customer/organizations/organizations.spec.js
```diff
@@ -0,0 +1,54 @@
...
+    it('SAAS-T160 organization is created automatically for admin', () => {
+      cy.oktaCreateUser(snAccount.admin1);
+
+      // login as admin and verify organization is created automatically
+      cy.loginByOktaApi(snAccount.admin1.email, snAccount.admin1.password);
+      openViewOrganizationPage();
+      verifyOrganizationName(snAccount.name);
+      cy.findByTestId('info-wrapper').should(
+        'have.text',
+        'We found your organization on Percona Customer Portal and used it',
```
nicolalamacchia, line 28: nit: I suggest to put this in a constant. The same applies to other messages.

cypress/support/ui/common.js
```diff
@@ -34,3 +34,26 @@ Cypress.Commands.add('checkPopUpMessage', (message) => {
       element.next('button').click();
     });
 });
+
+Cypress.Commands.add('getTable', { prevSubject: true }, (subj, options = {}) => {
+  // eslint-disable-next-line no-magic-numbers
+  if (subj.get().length > 1) throw new Error(`Selector "${subj.selector}" returned more than 1 element.`);
```
nicolalamacchia, line 40:
```suggestion
```suggestion
  if (subj.get().length) {
    throw new Error(`Selector "${subj.selector}" returned more than 1 element.`);
  }
```
```

**Review by nicolalamacchia — APPROVED** [review]

File: cypress/integration/customer/organizations/organizations.spec.js [hunk 4]
```diff
@@ -0,0 +1,60 @@
...
+      cy.loginByOktaApi(snAccount.admin1.email, snAccount.admin1.password);
+      openViewOrganizationPage();
+      verifyOrganizationName(snAccount.name);
+      cy.findByTestId('info-wrapper').should(
+        'have.text',
+        organizationPage.constants.messages.customerOrgFound,
+      );
+    });
+
+    it('SAAS-T207 SAAS-T218 technical user is added to members if org exists', () => {
```
peterSirotnak, line 34: More into testing comment than coding, but we should have a negative test. To try to add technical user to new org.

puneet0191 (dismissed):

peterSirotnak (dismissed):
LGTM

cypress/pages/gettingStarted.page.ts
```diff
@@ -0,0 +1,13 @@
+export const gettingStartedPage = {
+  constants: {
+    labels: {
+      addOrganization: 'Add Organization',
+      viewOrganization: 'View Organization',
+    },
+    messages: {
+      customerOrgFound: 'We found your organization on Percona Customer Portal and used it',
```
atymchuk, line 8: Can we reuse the messages defined in the app? Problem is when we change the message, we'll need to update here...

atymchuk: Apart from that, down below the same constant is defined as part of `organizationPage`. That one is used in the tests, this one probably isn't.

cypress/support/api/orgd_commands.js
```diff
@@ -0,0 +1,35 @@
+/// <reference types="cypress" />
+
+Cypress.Commands.add('apiCreateOrg', (accessToken, orgName = 'any') => {
```
atymchuk, line 3: The `any` looks like a typescript type and may create confusion when debugging. Maybe better call it something like `Test Organization`?

cypress/pages/gettingStarted.page.ts
yurkovychv: @atymchuk Great catch, bad copy-paste! thanks!
About reusing messages from a FE code: it is a bad practice. We should have our own messages on the tests side of things to catch changes and highlight them.

cypress/support/api/orgd_commands.js
yurkovychv: makes sense, will do

cypress/pages/gettingStarted.page.ts
atymchuk: Yes, well aware of the best practices guide, which is a good one if we want to keep ourselves busy :) In reality, we are not testing if the message is the same, we are most probably testing if it's there, which is pre-conditioned on the app logic. 

In other words, the fact that someone updated the title does not necessarily mean the app is broken. I'm thinking about the maintenance costs and want to ensure we code more tests vs just having fun updating the messages.

Anyway, I leave it up to you and the team, you know better )

atymchuk (approved):

cypress/pages/gettingStarted.page.ts
yurkovychv: I see your point and it totally makes sense. Will bring this topic to our QAA sync. Thanks for the suggestion @atymchuk  )

PR #198, opened 2022-03-24
Fix for update in okta widget and wait for dashboard to load

peterSirotnak:
There was an update in Okta widget where password field is only showed after email is verified, this is a fix for that. It is only temporary until new Login is imlemented.


cypress/integration/auth/login.spec.js
```diff
@@ -34,8 +34,9 @@ context('Login', () => {
 
   it('SAAS-T111 SAAS-T81 - should be able to login', () => {
     loginButton().click();
-    cy.get('#okta-signin-submit').should('be.visible');
-    cy.get('#okta-signin-username').type(newUser.email);
+    cy.get('#idp-discovery-username').should('be.visible');
```
yurkovychv, line 37:
```suggestion
    cy.get('#idp-discovery-username').should('be.visible').type(newUser.email);
```
You can use this way instead of having 2 lines with the same locator here

yurkovychv (dismissed):
one minor change

yurkovychv (dismissed):

beatahandzelova (approved):

PR #201, opened 2022-03-29
SAAS-848 simplify flow

nicolalamacchia:


src/components/GettingStarted/GettingStarted.tsx
```diff
@@ -27,11 +27,12 @@ export const GettingStartedPage: FC = () => {
   }, [dispatch, companyName]);
 
   return (
-    <PrivateLayout>
+    <>
+      <header className={styles.header}>{Messages.getStarted}</header>
```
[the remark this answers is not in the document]
nicolalamacchia: it is set as h2 in the style sheet

src/pages/Dashboard/Dashboard.tsx
```diff
@@ -29,11 +34,30 @@ export const DashboardPage: FC = () => {
...
+  }, [dispatch, orgId, orgs.length, companyName]);
+
+  useEffect(() => {
+    if (inventory == null && orgId) {
+      dispatch(getInventoryAction(orgId));
+    }
+  }, [dispatch, inventory, orgId]);
+
+  useEffect(() => {
+    setShowGettingStarted((!(companyName || orgId) || !inventory?.length) && !isPending);
```
fabio-silva, line 52: Not sure if I'm following this right:
- If `inventory` is empty, then show `GettingStarted`, as long as `isPending` is done;
- On `GettingStartedPmmSection` we have this:
    - `isTicked={!!inventory?.length}`
    -  `disabled={!inventory?.length}`

If we already know that inventory is empty when showing `GettingStartedPmmSection`, why are we retrieving it again there? 

Again, I might be missing something here.

fabio-silva (changes requested):

fabio-silva (dismissed):

src/components/GettingStarted/GettingStarted.tsx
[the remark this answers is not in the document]
TomasBlazek3PG: Not agree with this solution but its nothing that should block merging.

src/pages/Dashboard/Dashboard.tsx
TomasBlazek3PG: Could we write it differently for a better readability? some constants with good naming could help understand the code better.

nicolalamacchia: That would mean adding more `useState`, which I'd like to avoid. Those are the same variables used in other similar conditions in other parts of the code, so the logic is kind of self-explanatory.

Checking for `orgId` means checking if the org exists; checking for `companyName` means checking if the company is in Service Now; checking for `inventory` means checking if there are any linked PMM instances. That information is easily inferred by the context.

The doubt Fabio had was related to why we tick the last point in the list of getting started items to then just hide the component (effectively making useless the logic of `isTicked` in the last point).

TomasBlazek3PG: You dont need to use useState. Just to define the variables in the scope of the function. If you feel its not necessary then merge it but I would lean towards making it more readable.

TomasBlazek3PG (dismissed):

fabio-silva (approved):

TomasBlazek3PG (dismissed):

src/components/GettingStarted/GettingStarted.tsx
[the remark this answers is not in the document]
nicolalamacchia: the reason for this is that we don't have a global style that we control for `h*` tags, rather we use Grafana's themes, but we have to explicitly set something like `font-size: ${typography.size.h<n>}`. should we also use an `h<n>` tag, we should update both the style and the markup, this would be error-prone (I could change the type to - for example -`h3` from `h2` and forget about changing the markup which would be inconsistent) and less practical (admittedly sacrificing semantic markup for a more maintainable one). should we start maintaining our own themes, I agree with you that semantic markup would be better.

fabio-silva (dismissed):

fabio-silva (approved):

PR #205, opened 2022-04-06
SAAS-869 platform access token

nicolalamacchia:


src/core/utils.ts
```diff
@@ -38,3 +40,28 @@ export const omit = (keys: string[]) => (obj: AnObject) =>
...
+  } catch (e) {
+    console.error(e);
+    toast.error(Messages.platformAccessTokenNotFound);
+
+    return '';
+  }
+};
+
+export const copyToClipboard = async (text: string) => {
+  await navigator.clipboard.writeText(text);
```
fabio-silva, line 66: This will crash if not on HTTPS or localhost. Can we do a safety check, i.e. `if (navigator.clipboard)` ?

What we have on PMM:
```
if (navigator.clipboard) {
    navigator.clipboard
      .writeText(copiedText)
      .then(() => /**/);
} else {
    console.error("Clipboard API not accessible. Please make sure you're running over HTTPS or on localhost");
}
```

fabio-silva (changes requested):

src/core/utils.ts
fabio-silva: Source: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard#using_the_clipboard_api

`As the API is only available to [Secure Contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts), it cannot be used from a content script running on http:-pages, only https:-pages.`

nicolalamacchia: @fabio-silva I think it's pretty unlikely that people will access the platform not using https (same as accessing without JS). it might be worth implementing that check though, with just a generic "Clipboard API not accessible" error (`navigator.clipboard` could be not supported), will do that. I'm not implementing a fallback though, I don't think it's worth it (for now at least).

src/core/utils.ts
```diff
@@ -63,5 +63,9 @@ export const getPlatformAccessToken = (): string => {
 };
 
 export const copyToClipboard = async (text: string) => {
+  if (!navigator.clipboard) {
+    toast.error(Messages.clipboardNotAccessible);
+  }
```
fabio-silva, line 68: shouldn't we return from here? That was the point of the check, to avoid the app's crash, besides warning the user.

src/core/utils.ts
fabio-silva: Indeed, we don't need a fallback, it's purely a safety check to avoid a crash.

TomasBlazek3PG (approved):

src/core/utils.ts
nicolalamacchia: oops, fixed

fabio-silva (dismissed):

fabio-silva (approved):

PR #217, opened 2022-04-25
SAAS-19 add sentry integration

atymchuk:


it-percona:
@atymchuk could you please add environments https://docs.sentry.io/platforms/react-native/configuration/environments/ as an env variable so we can split dev from prod?

nicolalamacchia (approved):

PR #222, opened 2022-05-05
SAAS-900 Put ToS under social login

TomasBlazek3PG:


src/components/OktaSignInWidget/OktaSignInWidget.tsx
```diff
@@ -74,10 +74,45 @@ interface ContextProps {
...
+  if (widgetRef.querySelector('.tos-label')) {
     return;
   }
 
+  if (controller === 'registration') {
+    signUpToS(widgetRef);
+  }
+
+  if (controller === 'idp-discovery') {
+    socialToS(widgetRef);
```
fabio-silva, line 86: Please declare those functions before calling them here, for two reasons:

- Both are function expressions, they are not hoisted;
- Both are `const`. Even if they were hoisted, they would have been unitialized and a reference error might be thrown.

src/components/OktaSignInWidget/OktaSignInWidget.tsx
```diff
@@ -74,10 +74,45 @@ interface ContextProps {
...
+  }
+
+  const tosWrapper = document.createElement('div');
+
+  container.append(tosWrapper);
+  ReactDOM.render(
+    // eslint-disable-next-line jsx-a11y/label-has-associated-control
+    <p className='tos-label'>
+      {Messages.byContinuing}
+      <a href={TERMS_OF_SERVICE_URL} target='_blank' rel='noreferrer'>
```
fabio-silva, line 104: For compatibility reasons, please add to both anchors: `rel='noreferrer noopener'`

src/pages/Login/Login.styles.ts
```diff
@@ -117,6 +117,16 @@ export const getStyles = (theme: GrafanaTheme) => {
...
           }
         }
+
+        .social-auth-button {
+          padding-left: 50px;
+          padding-right: 0;
+          text-align: center;
+        }
+
+        .tos-label a, .tos-label a {
```
fabio-silva, line 127: Duplicate?

fabio-silva (changes requested):

src/pages/Login/Login.styles.ts
TomasBlazek3PG: Oh yeah it was suppose to be label a, p a but then I decided to use the class but forgot to get rid of the duplicate. Thanks.

fabio-silva (approved):

src/pages/Login/Login.styles.ts
fabio-silva: 👍

src/components/OktaSignInWidget/OktaSignInWidget.messages.ts
```diff
@@ -1,5 +1,8 @@
 export const Messages = {
   iHaveAgreed: 'I have read and agree to the ',
+  byContinuing:
+    // eslint-disable-next-line max-len
```
nicolalamacchia, line 4: is this needed?

src/components/OktaSignInWidget/OktaSignInWidget.tsx
```diff
@@ -69,15 +69,34 @@ const ToSCheckbox: FC<ToSCheckboxProps> = ({ submitBtn }) => {
...
-  if (controller !== 'registration' || widgetRef.querySelector('.tos-label')) {
+  if (!container) {
     return;
   }
 
+  const tosWrapper = document.createElement('div');
+
+  container.append(tosWrapper);
+  ReactDOM.render(
+    // eslint-disable-next-line jsx-a11y/label-has-associated-control
```
nicolalamacchia, line 83: is this needed?

src/components/OktaSignInWidget/OktaSignInWidget.tsx
```diff
@@ -100,6 +119,24 @@ const insertToS = ({ controller }: ContextProps, widgetRef: HTMLDivElement) => {
...
+    return;
+  }
+
+  if (controller === 'registration') {
+    signUpToS(widgetRef);
+  }
+
+  if (controller === 'idp-discovery') {
+    socialToS(widgetRef);
+  }
```
nicolalamacchia, line 137: either use a `else if` here or return in the first if

nicolalamacchia (changes requested):

nicolalamacchia (approved):

nicolalamacchia (approved):

PR #226, opened 2022-05-17
SAAS-914 Ticket Overview

TomasBlazek3PG:
https://files.slack.com/files-pri/T03FRR0BJ-F03GAFERYNL/image.png


package.json
```diff
@@ -14,6 +14,7 @@
   "publishConfig": {
     "access": "public"
   },
+  "proxy": "https://portal-dev.percona.com",
```
fabio-silva, line 17: I believe this is unintended

src/components/Charts/Pie/Pie.tsx
```diff
@@ -0,0 +1,52 @@
...
+        startAngle: p.startAngle,
+        endAngle: p.endAngle,
+      }),
+    );
+  }, [pie, size, width]);
+
+  return (
+    <div className={styles.wrapper}>
+      <svg width={size} height={size}>
+        {/* eslint-disable-next-line no-magic-numbers */}
```
fabio-silva, line 36: Let's disable the rule for the whole file instead

src/components/SupportTicketOverview/Legend/Legend.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const Legend: FC<ILegend> = ({ values }) => {
+  const styles = useStyles(getStyles);
+
+  return (
+    <div className={styles.legend}>
+      <div className={styles.row}>
+        {values.map((item) => (
+          <div className={styles.item}>
+            <div className={styles.colorNameNumber} key={item.department}>
```
fabio-silva, line 18: Should key be here or on the parent instead?

src/components/SupportTicketOverview/Legend/Legend.tsx
```diff
@@ -0,0 +1,39 @@
...
+              <span className={styles.color} style={{ backgroundColor: item.color }} />
+              {item.department}
+              <strong>{item.ammount}</strong>
+            </div>
+          </div>
+        ))}
+      </div>
+      <div className={styles.row}>
+        {values.map((item) => (
+          <div className={styles.item}>
```
fabio-silva, line 28: missing key

src/components/SupportTicketOverview/SupportTicketOverview.tsx
```diff
@@ -0,0 +1,121 @@
...
+    () =>
+      computedItems.map((item) => ({
+        name: item.department,
+        value: item.ammount,
+        color: item.color,
+      })),
+    [computedItems],
+  );
+
+  if ((!companyName || !orgId || !tickets) && !loading) return null;
```
fabio-silva, line 88: Missing brackets

src/components/SupportTicketOverview/SupportTicketOverview.tsx
```diff
@@ -0,0 +1,121 @@
...
+  const loading = useSelector(getOrgTicketsPending);
+  const companyName = useSelector(getUserCompanyName);
+  const styles = useStyles(getStyles);
+
+  const computedItems = useMemo(() => {
+    const categories = tickets.reduce((acc: ICategory[], ticket) => {
+      const alreadyInAcc = acc.find((category: ICategory) => category.department === ticket.department);
+
+      if (alreadyInAcc) {
+        // eslint-disable-next-line no-magic-numbers
```
fabio-silva, line 41: Same as the other file for disabling the rule

src/components/SupportTicketOverview/SupportTicketOverview.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <h4>{Messages.header}</h4>
+      <div className={styles.container}>
+        {loading ? (
+          <Skeleton height={230} />
+        ) : (
+          <div className={styles.containerPadding}>
+            <Legend values={computedItems} />
+            <Pie
+              values={pieValues}
+              size={200}
```
fabio-silva, line 101: Let's move these numbers to constants

src/components/SupportTicketOverview/SupportTicketOverview.tsx
```diff
@@ -0,0 +1,121 @@
...
+      return [
+        ...acc,
+        {
+          department: ticket.department,
+          ammount: 1,
+          color: '#fff',
+          types: [ticket.taskType],
+        },
+      ];
+    }, []);
```
fabio-silva, line 65: Just to be sure: this `computedItems` is supposed to be the same across renders?

fabio-silva (changes requested):
Besides all the other points, I realise we're mixing `className` with `styles`. That can be avoided, like this: https://github.com/percona-platform/grafana/blob/main/public/app/percona/integrated-alerting/components/Severity/Severity.styles.ts

`stylesFactory` is considered deprecated (it does memoization), but we can still return a style object based on props without adding separate styles.

src/components/SupportTicketOverview/SupportTicketOverview.tsx
TomasBlazek3PG: This is not the dependency array of useMemo. Its the accumulator of reduce function.

fabio-silva: Oh, right, my bad!

src/store/orgs/org.actions.ts
```diff
@@ -160,9 +160,10 @@ export const searchOrgsAction = createAsyncThunk<Organization[], void>(
   'ORGS:ORGANIZATIONS/SEARCH',
   async (_, { rejectWithValue }) => {
     try {
-      const { data } = await OrgAPI.searchOrgs();
+      /* const { data } = await OrgAPI.searchOrgs(); */
+      const data = await OrgAPI.searchOrgs();
 
-      return transformOrganizations(data.orgs);
+      return transformOrganizations(data.data.orgs);
```
fabio-silva, line 166: Can we remove this comment? Also, better do `result = await ...`, because `data.data` is a bit odd

fabio-silva (changes requested):

fabio-silva (approved):

nicolalamacchia (approved):

fabio-silva (approved):

PR #229, opened 2022-05-26
SAAS-917 fix customer org creation

nicolalamacchia:


package.json
```diff
@@ -18,6 +18,7 @@
     "type": "git",
     "url": "git+https://github.com/percona-platform/saas-ui.git"
   },
+  "proxy": "https://portal-dev.percona.com",
```
fabio-silva, line 21:
```suggestion
```

src/store/orgs/org.actions.ts
```diff
@@ -89,22 +89,26 @@ export const getOrganizationAction = createAsyncThunk<GetOrganizationResponse, s
...
     toast.info(Messages.fromCustomerPortal, { autoClose: false });
 
     return data;
-  } catch (err) {
-    displayAndLogError(err);
+  } catch (err: any) {
+    if (err?.response?.status) {
+      toast.info(Messages.adminFirst, { autoClose: false });
+    } else {
+      displayAndLogError(err);
```
fabio-silva, line 110: Is it possible to have an API error with no status??

fabio-silva (changes requested):

package.json
nicolalamacchia: Why? The proxy is only enabled in dev

src/store/orgs/org.actions.ts
nicolalamacchia: yes, because the type of the error is unknown. it's not realistic, but theoretically possible

package.json
fabio-silva: Works for me. But in that case our README should be updated, as it assumes the configuration is hidden by default:

`When developing locally, you'll want to route your API requests to a dev server. To do so, add the following key to your package.json:`

src/store/orgs/org.actions.ts
fabio-silva: 👍

fabio-silva (approved):

package.json
nicolalamacchia: you're right, done!

fabio-silva (approved):

fabio-silva (approved):

PR #232, opened 2022-06-01
SAAS-965 Fix linting rules

fabio-silva:
Fix lint rules


.eslintrc
```diff
@@ -33,7 +34,6 @@
     "no-restricted-imports": "error",
     "no-param-reassign": ["error", { "ignorePropertyModificationsForRegex": ["^acc$"] }],
     "react/jsx-curly-brace-presence": ["error"],
-    "react/jsx-curly-newline": ["error", "never"],
```
TomasBlazek3PG, line 36: I know @nicolalamacchia was very much against changing the rules. I am not sure about this particular one but I would ask him.

TomasBlazek3PG:
I think its fine but we need a consensus of us all.

fabio-silva:
> I think its fine but we need a consensus of us all.

Well, in my opinion, this repo is mostly used by you and @nicolalamacchia and the hook will therefore affect mostly you, PMM has this in place already for a long time. So, if you guys think it's too much, I can remove it and we can rely on the CI/CD, up to you.

.eslintrc
fabio-silva: The problem here is that this rule conflicts with prettier, which we use on all projects. So far we had no problem because prettier wasn't properly set up.

https://github.com/prettier/eslint-config-prettier/issues/126

TomasBlazek3PG (dismissed):

PR #245, opened 2022-06-15
SAAS-7 Run CI on push to dev branch

peterSirotnak:
Right now we do not have a simple way to trigger CI on dev branch, triggering it on push to dev makes a lot of sense since we need to deploy new env with every change pushed to it.


browseman (approved):
LGTM

nicolalamacchia (approved):

PR #280, opened 2022-08-22
SAAS-1077 update customer care contact address

nicolalamacchia:


src/pages/Dashboard/Contacts/Contacts.tsx
```diff
@@ -43,6 +44,10 @@ export const Contacts: FC = () => {
     [setIsEntilementsVisible],
   );
 
+  useEffect(() => {
```
fabio-silva, line 47: I think we don't need useState or useEffect for this. `helpEmail` is a derived state from the store.
This component will re-render every time `isCustomer` changes already, since it comes from the store.
So, I believe we can just do `const helpEmail = isCustomer ? CUSTOMER_HELP_EMAIL : FREE_USER_HELP_EMAIL`

src/pages/Dashboard/Contacts/Contacts.constants.ts
```diff
@@ -5,4 +5,5 @@ export const LINKS = {
   contact: 'https://www.percona.com/about-percona/contact',
 };
 
-export const HELP_EMAIL = 'portal-help@percona.com';
+export const FREE_USER_HELP_EMAIL = 'portal-help@percona.com';
```
TomasBlazek3PG, line 8: This is not a problem more of a question:
Whne we should add to src/core/constants.ts and when to local constants?

I think this might be reused in the future in other places.

TomasBlazek3PG (approved):

src/pages/Dashboard/Contacts/Contacts.constants.ts
nicolalamacchia: Mmm, not sure about that, currently the constants in the global constants file are kind of settings, aside from very few cases. I'm going to invoke YAGNI on this and say we can change the place of them, should we need those constants also in other places in the future.

fabio-silva (approved):

TomasBlazek3PG (approved):

PR #284, opened 2022-08-31
SAAS-1038: New activation page

TomasBlazek3PG:
image


src/core/api/endpoints.ts
```diff
@@ -7,6 +7,7 @@ export const ENDPOINTS = {
     SignOut: '/v1/auth/SignOut',
     SignUp: '/v1/auth/SignUp',
     UpdateProfile: '/v1/auth/UpdateProfile',
+    ActivateProfile: '/v1/auth/UpdateProfile',
```
fabio-silva, line 10: same as the one above?

src/pages/Activation/Activation.messages.ts
```diff
@@ -0,0 +1,21 @@
...
+    description: 'Create your account in order to get access to Percona Portal.',
+    name: 'Name',
+    surname: 'Surname',
+    yourPassword: 'Your password',
+    verifyYourPassword: 'Verify your password',
+  },
+  errors: {
+    password: {
+      lowerCase: 'Password needs to contain a lowercase letter',
+      upperCase: 'Password needs to contain a uppercase letter',
```
fabio-silva, line 14:
```suggestion
      upperCase: 'Password needs to contain an uppercase letter',
```

src/pages/Activation/Activation.tsx
```diff
@@ -0,0 +1,227 @@
...
+import { getStyles } from './Activation.styles';
+import { Messages } from './Activation.messages';
+
+const Activation = () => {
+  const styles = useStyles(getStyles);
+  const [activateUser, { status, isLoading, error: activationError }] = useActivateProfileMutation();
+  const history = useHistory();
+  const { search } = window.location;
+  const formRef: React.MutableRefObject<FormApi | null> = useRef(null);
+  const query = useMemo(() => new URLSearchParams(search), [search]);
```
fabio-silva, line 25: Do we really need `useMemo` for this? It seems an overkill, given that all those `q<name>` consts are derived from it and aren't changing when it changes as well.
We're using `search` as dependency. Is it gonna change unless we actually refresh the page, in this case? If so, `useMemo` brings no value.

src/pages/Activation/Activation.tsx
```diff
@@ -0,0 +1,227 @@
...
+                          <span className={styles.infoIcon}>&#9432;</span>Your password should:
+                        </p>
+                        <ul className={styles.passwordList}>
+                          <li>Contain at least 10 characters</li>
+                          <li>Contain a lowercase letter</li>
+                          <li>Contain an uppercase letter</li>
+                          <li>Contain a number</li>
+                          {/* <li>Not contain parts of your username</li> */}
+                          <li>Not include your first or last name</li>
+                          <li>Not be the one of your last 4 passwords</li>
```
fabio-silva, line 159:
```suggestion
                          <li>Not be one of your last 4 passwords</li>
```

fabio-silva (changes requested):

fabio-silva (dismissed):

MaksimMalakhovSimbirsoft (approved):

MaksimMalakhovSimbirsoft (approved):

PR #285, opened 2022-08-31
SAAS-1075 Fix ui-tests 

oksana-grishchenko:
The error expectations are updated according to the new validation approach:
The `org.id` is checked first whether it represents a valid uuid, so the test case that uses "0" as the `org.id` should expect another error, which is implemented in this PR  

SAAS-1075 

- [ ] https://github.com/percona-platform/api-tests/pull/162
- [ ] https://github.com/percona-platform/orgd/pull/206


peterSirotnak (approved):
LGTM

PR #296, opened 2022-09-21
SAAS-1121: Add message email

MaksimMalakhovSimbirsoft:
Screen sent


src/core/okta.ts
```diff
@@ -38,6 +38,7 @@ export const authConfig = {
       'registration.form.submit': Messages.signUpButton,
       'socialauth.google.label': Messages.signInGoogle,
       'socialauth.github.label': Messages.signInGithub,
+      'registration.complete.confirm.text': 'To finish signing in, check your registration email',
```
nicolalamacchia, line 41: put this in the messages file

src/components/OktaSignInWidget/OktaSignInWidget.messages.ts
```diff
@@ -8,4 +8,7 @@ export const Messages = {
   andPerconas: ' and Percona ',
   privacyPolicy: 'Privacy Policy',
   perconaPrivacyPolicy: 'Percona Privacy Policy',
+  verificationEmailSent: (email: string) => `To finish signing in, check your email ${email}.`,
+  iConsent:
```
nicolalamacchia, line 12: nit: I personally don't like "iConsent", I'd perfer just "consent"

src/components/OktaSignInWidget/OktaSignInWidget.tsx
```diff
@@ -204,6 +204,7 @@ export const OktaSignInWidget = ({
 }) => {
   const styles = useStyles(getStyles);
   const widgetRef = useRef<HTMLDivElement>(null);
+  const [email, setEmail] = useState<string>('');
```
nicolalamacchia, line 207: I don't understand why we need an email state here. I see it's only being used to trigger the same effect that sets it. Please explain this.

nicolalamacchia (changes requested):

PR #316, opened 2022-10-31
SAAS-1153 migrate tables to MUI

nicolalamacchia:


MaksimMalakhovSimbirsoft (approved):

src/core/utils.ts
```diff
@@ -71,3 +73,24 @@ export const copyToClipboard = async (text: string) => {
 
   await navigator.clipboard.writeText(text);
 };
+
+function stylesFactory<ResultFn extends (this: any, ...newArgs: any[]) => ReturnType<ResultFn>>(
```
TomasBlazek3PG, line 77: Can you rewrite it as `const styleFactory = () => {} `? Just so its all the same way? The same with useStyles.

nicolalamacchia: good catch, done

PR #327, opened 2022-11-10
SAAS-1160: Activation Page MUI Migration

TomasBlazek3PG:
image


src/pages/Activation/Activation.styles.ts
```diff
@@ -1,52 +1,73 @@
-import { GrafanaTheme } from '@grafana/data';
+import { Theme } from '@mui/material';
 import { css } from 'emotion';
 
-export const getStyles = ({ colors, spacing, typography, palette }: GrafanaTheme) => ({
+const bgColor = '#f1f5f9';
```
nicolalamacchia, line 4: I would remove the BG, to make this more uniform to with the design

src/pages/Activation/Activation.tsx
```diff
@@ -108,112 +109,145 @@ const Activation = () => {
...
-                        </a>
-                        .&nbsp;*
-                      </>
-                    }
-                    name="tos"
-                    fieldClassName={styles.checkbox}
-                    required
-                  />
-                </div>
+                          <span className={styles.infoIcon}>&#9432;</span>
```
nicolalamacchia, line 203: let's use an icon for this, that depends on how the font is rendering. it's not readable and it's inconsistent with what we do elsewhere.

nicolalamacchia (approved):
Approving this to speed up the process, but please make those small changes.

PR #331, opened 2022-11-11
SAAS-1161 migrate the login page to MUI

nicolalamacchia:


src/pages/Login/Login.tsx
```diff
@@ -1,10 +1,10 @@
 import React, { FC, useCallback } from 'react';
-import { useStyles } from '@grafana/ui';
 import { OktaSignInWidget } from 'components/OktaSignInWidget';
 import { authConfig } from 'core';
+import { useStyles } from 'core/utils';
 import { useOktaAuth } from '@okta/okta-react';
 import logoImg from 'assets/percona-platform-logo.svg';
-import { logger } from '@percona/platform-core';
+import * as logger from 'core/logger';
```
fabio-silva, line 7: This is something we can probably avoid.
Check https://github.com/percona/grafana-dashboards/blob/main/pmm-app/src/shared/core/logger.ts

fabio-silva (changes requested):

src/pages/Login/Login.styles.ts
```diff
@@ -1,183 +1,178 @@
...
+      text-decoration: none;
+      align-items: center;
+      line-height: 30px;
+      vertical-align: middle;
       display: flex;
 
       img {
-        width: 100%;
-        max-width: 350px;
+        margin-left: 7px;
```
TomasBlazek3PG, line 47: please use spacing

src/pages/Login/Login.styles.ts
```diff
@@ -1,183 +1,178 @@
...
+        margin-right: ${spacing(3)};
       }
-    `,
-    texts: css`
-      color: ${palette.white};
+    }
+
+    h2 {
+      font-size: 32px;
```
TomasBlazek3PG, line 53: please use predefined sizes

src/pages/Login/Login.tsx
```diff
@@ -28,12 +28,12 @@ export const LoginPage: FC = () => {
...
-      <div className={styles.leftSide}>
-        <div className={styles.logo}>
+    <main data-testid="login-page-container" className={styles.container}>
+      <div data-testid="login-page-left-side-wrapper" className={styles.leftSide}>
+        <div data-testid="login-page-logo" className={styles.logo}>
           <img src={logoImg} alt={Messages.logoAlt} />
         </div>
-        <div className={styles.texts}>
+        <div data-testid="login-page-title" className={styles.texts}>
           <h2>{Messages.title}</h2>
```
TomasBlazek3PG, line 37: Would use Typography component

TomasBlazek3PG (changes requested):
Using of the predefined sizes is all over the styles so I didnt mention it in every place.

src/pages/Login/Login.styles.ts
nicolalamacchia: this is a nit and that's from the existing code, out of the scope of this PR. Also this is likely going to go away after the redesign.

src/pages/Login/Login.styles.ts
nicolalamacchia: this is a nit and that's from the existing code, out of the scope of this PR. Also this is likely going to go away after the redesign.

src/pages/Login/Login.tsx
nicolalamacchia: good point, thanks! done

src/pages/Login/Login.tsx
nicolalamacchia: good point, done

PR #332, opened 2022-11-16
SAAS-1176: Handle Invalid Activation Links

TomasBlazek3PG:
image


nicolalamacchia:
@TomasBlazek3PG the button here looks like it's still from Grafana

nicolalamacchia (approved):

PR #335, opened 2022-11-22
SAAS-1211 migrate getting started to MUI

nicolalamacchia:
image
image
image


src/components/GettingStarted/GettingStarted.tsx
```diff
@@ -10,7 +11,9 @@ export const GettingStarted: FC = () => {
 
   return (
     <>
-      <header className={styles.header}>{Messages.getStarted}</header>
+      <Typography variant="h5" className={styles.header}>
```
TomasBlazek3PG, line 14: SHould we use variant or component? I guess component?

src/components/GettingStarted/GettingStartedOrgSection/GettingStartedOrgSection.tsx
```diff
@@ -18,7 +19,7 @@ export const GettingStartedOrgSection: FC = () => {
     <GettingStartedSection
       description={Messages.createOrganizationDescription}
       title={Messages.createOrganizationTitle}
-      linkIcon={hasOrg ? undefined : 'plus-circle'}
+      linkIcon={hasOrg ? undefined : <Add />}
```
TomasBlazek3PG, line 22: maybe hasOrg && <Add />?

src/components/GettingStarted/GettingStartedSection/GettingStartedSection.styles.ts
```diff
@@ -1,77 +1,72 @@
...
-    descriptionWrapper: css`
-      align-items: center;
-      display: flex;
-      margin-left: calc(${spacing.lg} + ${headerImageSize});
-    `,
-    header: css`
-      align-items: center;
-      display: flex;
-      font-size: ${typography.heading.h4};
+    > h5 {
```
TomasBlazek3PG, line 20: Why we dont target it dirrectly?

src/components/GettingStarted/GettingStarted.tsx
nicolalamacchia: we should use `variant` as per documentation

src/components/GettingStarted/GettingStartedOrgSection/GettingStartedOrgSection.tsx
nicolalamacchia: no, this is correct, `false && 'Test'` would result in `false`, we need undefined

src/components/GettingStarted/GettingStartedSection/GettingStartedSection.styles.ts
nicolalamacchia: fixed

TomasBlazek3PG (approved):

PR #338, opened 2022-11-23
SAAS-1155: Migrate Dashboard to MUI

TomasBlazek3PG:
image
(The left menu is fine, its just the screenshot tool that cut it)
image


src/pages/Dashboard/Contacts/AccountInfo/AccountInfo.styles.ts
```diff
@@ -1,33 +1,34 @@
...
     flex-wrap: wrap;
-    margin-top: ${spacing.base * 7}px;
+    margin-top: ${spacing(3)};
   `,
   card: css`
-    color: ${colors.text};
     flex: 0 1 35%;
-    background-color: ${colors.bg3};
+    padding: ${spacing(3)};
+    background-color: ${palette.mode === 'light' ? blueGrey[50] : blueGrey[900]};
```
nicolalamacchia, line 14: let's use just one color and let's not switch it based on the theme. that's something MUI does automatically

src/pages/Dashboard/Contacts/Advisor/Advisor.styles.ts
```diff
@@ -1,20 +1,20 @@
...
 
-export const getStyles = ({ spacing, palette }: GrafanaTheme) => ({
+export const getStyles = ({ spacing, palette }: Theme) => ({
   wrapper: css`
     align-items: center;
     display: flex;
   `,
   label: css`
-    width: 110px;
+    width: 150px;
```
nicolalamacchia, line 10: use spacing(x)

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
```diff
@@ -1,73 +1,63 @@
...
-        div[class$="panel-container"]:nth-child(${i + NEXT_CHILD}) {
-          div[class$="collapse__header-label"] {
-            width: 100%;
+        `.EntitlementAccordions:nth-child(${i + NEXT_CHILD}) {
+          .MuiAccordionSummary-content {
+            position: relative;
+          }
 
-            &::after {
+          .MuiAccordionSummary-content::after {
```
nicolalamacchia, line 26: I wouldn't do this, let's just use basic MUI style

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
```diff
@@ -1,73 +1,63 @@
...
     ${entitlements.map(
       ({ end_date }, i) =>
-        `div[data-testid="modal-content"] {
-        div[class$="panel-container"]:nth-child(${i + NEXT_CHILD}) {
-          div[class$="collapse__header-label"] {
-            width: 100%;
+        `.EntitlementAccordions:nth-child(${i + NEXT_CHILD}) {
+          .MuiAccordionSummary-content {
+            position: relative;
+          }
```
nicolalamacchia, line 24: what is this for?

nicolalamacchia (approved):

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
TomasBlazek3PG: to make the absolute ::after element use .MuiAccordionSummary-content as containing box for its positioning.

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
TomasBlazek3PG: This is targeting subelement. It must be done this way.

src/pages/Dashboard/Contacts/AccountInfo/AccountInfo.styles.ts
TomasBlazek3PG: when I put blueGrey[50] it doesnt change with modes

src/pages/Dashboard/Contacts/Advisor/Advisor.styles.ts
TomasBlazek3PG: Are you sure? we use px everywhere. Spacing is best use for margins and paddings.

src/pages/Dashboard/Contacts/AccountInfo/AccountInfo.styles.ts
nicolalamacchia: that doesn't, but the other palette colors do, like `primary.main` or the color for borders, which are the ones we should use. when in doubt, leave the background white.

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
nicolalamacchia: my point is that maybe we shouldn't change it, let's use MUI basic style as much as possible

src/pages/Dashboard/Contacts/EntitlementsModal/EntitlementsModal.styles.ts
nicolalamacchia: it's not clear to me why it's needed, why can't we just use the basic accordion as it is?

src/pages/Dashboard/Contacts/Advisor/Advisor.styles.ts
nicolalamacchia: usually yes, but not necessarily. we don't have a grid system yet, so it's ok to replace it with spacing for now. either way it's a fixed width, which we should change. consider this a nit, up to you.

PR #349, opened 2022-12-14
SAAS-1192: Activity Log

TomasBlazek3PG:
image
image


nicolalamacchia (approved):

src/core/api/events.service.ts
```diff
@@ -0,0 +1,41 @@
+import { createApi, fetchBaseQuery, BaseQueryFn, FetchArgs } from '@reduxjs/toolkit/query/react';
+import { ENDPOINTS } from 'core/api';
+import { EventsSearchRequest, EventsSearchResponse } from 'core/api/types';
+import { oktaAuth } from 'core';
+import { EditProfileError } from '../../components/MarketingBanner/MarketingBanner.types';
```
nicolalamacchia, line 5: please use absolute path, `import { EditProfileError } from 'components'` should be enough

src/pages/ManageOrganization/ActivityLog/ActivityLog.tsx
```diff
@@ -0,0 +1,175 @@
...
+        }))[0] || {
+          field_name: 'timestamp',
+          order: 1,
+        },
+      },
+    });
+  }, [pageSize, pageIndex, timeframe, sortModel, eventSearch, orgId]);
+
+  const handlePageChange = (newPageIndex: number) => setPageIndex(newPageIndex);
+  const handlePageSizeChange = (newPageSize: number) => setPageSize(newPageSize);
```
nicolalamacchia, line 66: shouldn't this be handled automatically by DataGrid?

src/pages/ManageOrganization/ActivityLog/ActivityLog.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const handleSelectTimeframe = (e: SelectChangeEvent<TimestampRangeType>) => {
+    setPageIndex(0);
+    setTimeframe(e.target.value as TimestampRangeType);
+  };
+  const handleSortModelChange = (model: GridSortModel) => setSortModel(model);
```
nicolalamacchia, line 71: why is this needed?

src/pages/ManageOrganization/ActivityLog/ActivityLog.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const renderDetailCell = ({ row }: { row: AuditEvent }) => {
+    const content = getDetailsText(row);
+
+    return (
+      <div title={content} className={styles.scrollableCell}>
+        {content}
+      </div>
+    );
+  };
```
nicolalamacchia, line 104: `render...` function are a code smell, you should create separate components for those

src/pages/ManageOrganization/ActivityLog/ActivityLog.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  const getDetailsText = (row: AuditEvent) => {
+    if ([EventType.LOG_IN, EventType.LOG_OUT, EventType.CHANGE_PASSWORD].includes(row.action_type)) {
+      return `User ${row.okta_user_id} ${eventTypeMapping[row.action_type]}`;
+    }
+
+    return `User ${row.okta_user_id} ${eventTypeMapping[row.action_type]} ${getListOfAffectedObjects(
+      row.objects,
+    )}`;
+  };
```
nicolalamacchia, line 88: what do these do?

nicolalamacchia (changes requested):

src/pages/ManageOrganization/ActivityLog/ActivityLog.tsx
TomasBlazek3PG: This creates human readable details column text like "User xxx updated xxxx,xxxx,xxxx instance"

src/pages/ManageOrganization/ActivityLog/ActivityLog.tsx
TomasBlazek3PG: This is setting the sort model (which way and which column should be sorted). It is then used in the request to BE.

src/pages/ManageOrganization/ActivityLog/ActivityLog.tsx
TomasBlazek3PG: We need to know what is the size so we can add this info to the request to BE

nicolalamacchia: right, I forgot it's paginated on the BE

nicolalamacchia (changes requested):
Looks good, but tests are missing

nicolalamacchia (dismissed):

nicolalamacchia (approved):

PR #356, opened 2023-01-05
SAAS-1183 - migrate logs to json format

browseman:
Modify the docker build process to apply new JSON output for NGINX server so it's easier to parse it in ElasitcCloud Kibana and Papertrail


Dockerfile
```diff
@@ -1,7 +1,12 @@
-FROM nginx:1.21.6
+FROM nginx:stable
 
-# Remove the default config
-RUN rm /etc/nginx/conf.d/default.conf
+# Backup the default config files
+RUN mv -v /etc/nginx/nginx.conf /etc/nginx/nginx.conf_
+RUN mv -v /etc/nginx/conf.d/default.conf /etc/nginx/conf.d/default.conf_
```
guillermoelia, line 5: Should we keep the original files in the image?

browseman: Decided to keep them just for the sake of any future needs to see how our custom config is diverged from the default one.
If we delete them, it will be more difficult (more involved) to get the particular defaults.

nginx_daemon.conf
```diff
@@ -0,0 +1,54 @@
...
+
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
+
+    keepalive_timeout  65;
+
+    #gzip  on;
```
nicolalamacchia, line 51: This can be removed

nginx_daemon.conf
```diff
@@ -0,0 +1,54 @@
...
+    "http_x_b3_traceid": "$http_x_b3_traceid",
+    "upstream_addr": "$upstream_addr",
+    "source": "nginx"
+}';
+
+
+    access_log  /var/log/nginx/access.log  main;
+
+    sendfile        on;
+    #tcp_nopush     on;
```
nicolalamacchia, line 47: Why is this commented out?

nicolalamacchia (approved):

nginx_daemon.conf
browseman: It was like that as always, this is the default config. The only thing new here is this block:

`    # JSON Extended Log format
    log_format  json  escape=json '{
    "time": "$msec",
    "time_iso8601": "$time_iso8601",
    "remote_addr": "$remote_addr",
    "remote_user": "$remote_user",
    "request": "$request",
    "status": "$status",
    "request_uri": "$request_uri",
    "request_method": "$request_method",
    "request_time": "$request_time",
    "request_length": "$request_length",
    "body_bytes_sent": $body_bytes_sent,
    "http_host": "$http_host",
    "http_referer": "$http_referer",
    "http_user_agent": "$http_user_agent",
    "http_x_forwarded_for": "$http_x_forwarded_for",
    "http_x_b3_traceid": "$http_x_b3_traceid",
    "upstream_addr": "$upstream_addr",
    "source": "nginx"
}';
`

nginx_daemon.conf
browseman: Same here, this is the default config. I rather keep it as is because:
- it was set like that prior to my changes
- it will be easier to track any future changes that we need to apply if the default config is changed.

Dockerfile
browseman: @guillermoelia does that make sense for you? May be you have some other thought about it?

guillermoelia: Yes that makes sense, just wanted to understand the reasoning

guillermoelia (approved):
LGTM

guillermoelia (approved):
LGTM, log readability in stdout much better

PR #363, opened 2023-01-24
Saas 1232 update marketing optin message

YashSartanpara1:
This PR is to update marketing consent message displayed on Signin and Signup page, This change will update the message to take passive consent hence the marketing checkbox from signup page will be removed.


TomasBlazek3PG (approved):

src/components/OktaSignInWidget/ToSCheckbox.styles.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Theme } from '@mui/material';
+
+export const getStyles = ({ spacing }: Theme) => ({
+  marketingLabel: css`
+    margin-bottom: ${spacing(1)} !important;
+    font-size: 0.75rem !important;
+  `,
+  tosLink: css`
+    text-decoration: none !important;
+    color: #007dc1 !important;
```
nicolalamacchia, line 11: let's use colors from the theme whenever possible. same for font sizes.

nicolalamacchia (changes requested):

nicolalamacchia (approved):

src/components/OktaSignInWidget/ToSCheckbox.styles.ts
YashSartanpara1: Sure.

PR #373, opened 2023-01-31
SAAS-1257 enhance profile page

YashSartanpara1:
This change is to update profile page design to Card view and temporarily remove form. Copy token functionality will work as it is. Ui-tests are skipped temporarily.


src/pages/Profile/Profile.messages.ts
```diff
@@ -1,14 +1,15 @@
 export const Messages = {
-  editProfile: 'Edit profile',
-  emailLabel: 'Email',
+  editProfile: 'Edit Profile',
+  emailLabel: 'Email: ',
```
nicolalamacchia, line 3: Let's use CSS for spacing instead of a space in the text

src/pages/Profile/Profile.styles.ts
```diff
@@ -1,5 +1,35 @@
...
+  `,
+  cardDescription: css`
+    padding: ${spacing(2)} 0 ${spacing(2)} 0;
+    margin: 0;
+  `,
+  cardHeader: css`
+    padding-top: ${spacing(2)};
+    margin: 0;
+  `,
+});
```
nicolalamacchia, line 31: is all of this CSS needed? I think the default card CSS is good, plus we should use the theme as much as possible CSS (so either we change the style for the buttons there of we keep the one we have).

src/pages/Profile/Profile.tsx
```diff
@@ -49,87 +34,70 @@ export const ProfilePage: FC = () => {
...
-                  sx={styles.submitButton}
-                  disabled={!valid || pending || pristine}
-                >
-                  {pending ? Messages.loading : Messages.save}
-                </Button>
-              </Box>
-            </Box>
-          )}
-        </Form>
+        <Grid container spacing={2}>
```
nicolalamacchia, line 37: you can simplify this structure using CSS instead of creating an HTML tree that becomes difficult to read

nicolalamacchia (changes requested):

src/pages/Profile/Profile.styles.ts
YashSartanpara1: just found that theme change is there in dev only, ill get the updated theme from there and remove these styles.

src/pages/Profile/Profile.messages.ts
YashSartanpara1: Sure.

src/pages/Profile/Profile.tsx
YashSartanpara1: Sure.

src/pages/Profile/Profile.tsx
```diff
@@ -34,103 +26,48 @@ export const ProfilePage: FC = () => {
...
-        </Form>
-      </Box>
+      <main className={styles.wrapper}>
+        <Card className={styles.card} elevation={4}>
+          <div className={styles.cardContent}>
+            <Typography variant="h5" data-testid="profile-details-header">
+              {Messages.myProfileLabel}
+            </Typography>
+            <p>
+              <Typography>
```
nicolalamacchia, line 43: Why do we need `Typography` here?

nicolalamacchia (approved):

src/pages/Profile/Profile.tsx
YashSartanpara1: Removed. instead used css to manage spacing.

nicolalamacchia (approved):

PR #381, opened 2023-02-08
Bump ubuntu version on runner to latest

guillermoelia:
Lets bump ubuntu to latest and check what breaks.


nicolalamacchia:
Solved in #393 using 22.04

PR #383, opened 2023-02-09
Saas 1258 user menu update

YashSartanpara1:
This change is to Update User menu to include User's full name and email. Theme change (dark mode) button will be removed from header and will be moved to User menu as Switch.


src/components/AppHeader/AppHeader.styles.ts
```diff
@@ -1,11 +1,21 @@
...
+    width: calc(100% - ${SIDE_MENU_WIDTH}px) !important;
+    z-index: 1000 !important;
+  `,
+  menu: css`
+    z-index: 1000;
+  `,
+  switch: css`
+    padding-top: 0 !important;
+    padding-bottom: 0 !important;
+    justify-content: space-between !important;
```
nicolalamacchia, line 24: I think we should find a better solution to these `!important`, they're spreading in the codebase. But let's discuss separately

nicolalamacchia (dismissed):

src/components/AppHeader/AppHeader.styles.ts
YashSartanpara1: Agree.

nicolalamacchia (approved):

PR #397, opened 2023-03-03
SASS-1101 Percona Portal request data when user starts new session after pause

doracretu3pillar:
Moved the check for state.email from store.ts to an axios interceptor in api.service.ts


nicolalamacchia (changes requested):

src/components/SecureRoute/SecureRoute.tsx
```diff
@@ -12,8 +16,27 @@ export const SecureRoute: FC<
...
+  const dispatch = useDispatch();
+  const userEmail = useSelector(getUserEmail);
+
+  useEffect(() => {
+    const getUser = async () => {
+      const data: UserClaims = await oktaAuth.getUser();
+
+      dispatch(
+        updateUserInfoAction({ email: data?.email, firstName: data?.first_name, lastName: data?.last_name }),
+      );
```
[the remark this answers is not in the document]
doracretu3pillar: fixed

src/components/MarketingBanner/MarketingBanner.tsx
```diff
@@ -1,15 +1,21 @@
...
 import { displayAndLogError, useStyles } from 'core';
 import { QueryStatus } from '@reduxjs/toolkit/dist/query';
 import { LoadingButton } from '@mui/lab';
+import { UserClaims } from '@okta/okta-auth-js';
 import { getStyles } from './MarketingBanner.styles';
 import { Messages } from './MarketingBanner.messages';
 import { useEditProfileMutation } from '../../core/api/auth.service';
 
-export const MarketingBanner: FC = ({ children }) => {
+interface IMarketingBanner {
```
nicolalamacchia, line 13:
```suggestion
interface MarketingBannerProps {
```

Please don't prepend interfaces with `I`

src/components/SecureRoute/SecureRoute.tsx
```diff
@@ -12,8 +15,37 @@ export const SecureRoute: FC<
...
     HTMLAttributes<HTMLDivElement>
-> = ({ children, ...props }) => (
-  <OktaSecureRoute {...props}>
-    <MarketingBanner>{children}</MarketingBanner>
-  </OktaSecureRoute>
-);
+> = ({ children, ...props }) => {
+  const dispatch = useDispatch();
+  const userEmail = useSelector(getUserEmail);
+  const [userInfo, setUserInfo] = useState({ sub: '' });
```
nicolalamacchia, line 21: why do we need the `{ sub: '' }` part?

nicolalamacchia (changes requested):

src/components/SecureRoute/SecureRoute.tsx
doracretu3pillar: You're right, first time the userInfo is actually undefined.
Please check again.

nicolalamacchia (approved):

src/components/MarketingBanner/MarketingBanner.test.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+import { AuthState } from '../../store/types';
+import { testRender } from '../../testUtils';
```
nicolalamacchia, line 4: Use absolute imports

src/components/MarketingBanner/MarketingBanner.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    authState: { isAuthenticated: true },
+  }),
+}));
+
+describe('Marketing Banner', () => {
+  const setup = (state: Partial<{ auth: Partial<AuthState> }>) =>
+    testRender(<MarketingBanner userInfo={state.auth} />);
+
+  it('should not render the marketing banner when email is empty', async () => {
+    // arrange
```
nicolalamacchia, line 25: these comments are not needed, it's a good practice not to comment what's already clear from the code

src/components/SecureRoute/SecureRoute.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }),
+  withOktaAuth: (x: any) => x,
+  Security: () => <div />,
+  SecureRoute: () => <div />,
+  LoginCallback: () => <div />,
+}));
+
+describe('Secure Route', () => {
+  const setup = (state: Partial<{ auth: Partial<AuthState> }>) =>
+    testRender(<SecureRoute />, {
```
nicolalamacchia, line 35: Why not using the `TestContainer` we already have, which does something similar?

nicolalamacchia (changes requested):

src/components/SecureRoute/SecureRoute.test.tsx
doracretu3pillar: TestContainer returns the JSX to be rendered and export the store that doesn't the `dispatch` mocked.
I've tried with TestContainer, I've mocked the store.dispatch but it doesn't get called so the test doesn't pass (I have an explanation for that).

testUtils.jsx returns the already rendered JSX and the store . Using the store you can access `dispatch` and `getState()`. So a wider range of information can be tested.
Cannot get rid of TestContainer or refactor it without affecting all the tests. 




https://redux.js.org/usage/writing-tests
if you search `renderWithProviders()` , the Wrapper is actually the TestContainer.

nicolalamacchia (approved):

nicolalamacchia (approved):

PR #405, opened 2023-03-16
Saas-1304 DBaaS Landing Page 

doracretu3pillar:


src/components/Root/Main.tsx
```diff
@@ -11,6 +11,7 @@ import {
   K8sClusterCreationPage,
```
nicolalamacchia, line 11: You should remove `K8sClusterCreationPage` if you don't want to reuse it

src/core/routes.ts
```diff
@@ -17,6 +17,7 @@ export const Routes = {
   welcome: '/welcome',
   dashboard: '/dashboard',
   kubernetes: '/kubernetes',
```
nicolalamacchia, line 19: we no longer need this if we're using a different URL

src/core/routes.ts
```diff
@@ -25,4 +26,5 @@ export const RouteNames = {
   [Routes.instances]: 'PMM Instances',
   [Routes.kubernetes]: 'Free Kubernetes Cluster',
```
nicolalamacchia, line 27: Remove this as well

nicolalamacchia (approved):

src/components/Root/Main.tsx
doracretu3pillar: removed

src/core/routes.ts
doracretu3pillar: removed

src/core/routes.ts
doracretu3pillar: removed

